PR #15, opened 2017-12-15
Processing now working as expected with object-oriented processor class

monocongo:
Both climate divisions and nClimGrid appear to be computing as expected using these new codes. Errors did occur on Windows due to numba compilation issues, in these cases a comment out of the numba annotation seems to fix the issue, none others are known.


monocongo:
Removing this, replacing with new pull request.

PR #25, opened 2017-12-27
Merge several changes for reorganization and divisions/grid processing

monocongo:


monocongo:
Now with reorganized folders and with nClimDiv processor working well from ASCII to indices and comparison plots/differences histograms. Corresponding nClimDiv/PRISM processor underway/testing, not fully fleshed out yet but included nevertheless. Still challenged by absolute imports, resorting to a context hack for test classes, such a remedial approach may be necessary for other modules as well as they're discovered.

PR #34, opened 2017-12-30
Removed need for context.py after directory restructure

synflyn28:
James,

I re-arranged the code so that the context.py file or messing with sys.path is no longer needed. This is an issue that I've struggled with in Python many times and isn't an obvious issue to fix. For your Python project you not only need to have the project root directory be the name of the package you want to create, but you also need to create a **sub-package underneath the project root with the same name as the package** where you will store the code you want to import. So in other words the project directory tree I changed it to is:
indices_python
|
|_ _ _ _ _ _indices_python/: this is where you want to store all of your importable modules
|----tests/
|----setup.py
|----etc...

This seems really redundant (and it is in my opinion) but that is how Python's packaging does things. I ran your tests on my end and everything seems ready to go!


monocongo:
Justin I *really* appreciate this, as it's been vexing me for some time. Every article I read about this didn't explain things nearly as well or succinctly as you did above. So helpful, thanks!

synflyn28:
Sure thing James, anytime!

PR #35, opened 2018-01-02
Code with comparison script for climate divisions, reorganization into directories, absolute imports

monocongo:
Comparison analysis branch builds and contains code for ingest to NetCDF, indices processing, and comparison analysis plots.


monocongo:
Works as expected, tests/builds OK

Issue #50, opened 2018-01-23
Clean up project based on suggestions from PyAOS mailing list


monocongo:
- [ ] Remove \_\_init\_\_.py files where not needed
- [x] Consolidate a single top-level .gitignore file

PR #51, opened 2018-01-24
Issue50gitignore

kikocorreoso:
Partial fix for #50.


monocongo (approved):
lots of goodies added, much more complete, thanks!

PR #53, opened 2018-01-24
remove nose as a dependency as test suite is made with unittest

kikocorreoso:


monocongo:
The build fails due to missing nose requirement (several falures, single example below taken from Travis CI):

> ======================================================================
> ERROR: test_sign_change (tests.test_utils.UtilsTestCase)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/travis/build/monocongo/indices_python/tests/test_utils.py", line 131, in test_sign_change
>     np.array([1., 2., 3.]))
>   File "/home/travis/miniconda/envs/indices_python/lib/python3.6/site-packages/numpy/testing/utils.py", line 1189, in assert_raises
>     nose = import_nose()
>   File "/home/travis/miniconda/envs/indices_python/lib/python3.6/site-packages/numpy/testing/utils.py", line 70, in import_nose
>     raise ImportError(msg)
> ImportError: Need nose >= 1.0.0 for tests - see http://nose.readthedocs.io

kikocorreoso:
Strange. I've created a virtualenv and installed just numpy, scipy, pycurl, numba and pandas (and indices_python) and after `python -m unittest -v` I get:

```python
----------------------------------------------------------------------
Ran 26 tests in 18.205s

OK
```

I have to check this in detail.

kikocorreoso:
Maybe the error is related with this line:

https://github.com/monocongo/indices_python/blob/master/.travis.yml#L23

Try adding `=` after `--source`, i.e., replace this:

`coverage run --source indices_python -m unittest tests/test_*.py`

with:

`coverage run --source=indices_python -m unittest tests/test_*.py`

Source: https://coverage.readthedocs.io/en/coverage-4.4.2/cmd.html#execution

monocongo:
I have tested adding the equals sign as you've suggested, it appears to work fine now with that added. Can you please add this to the travis.yml file on your branch to merge (rm_nose_dep) so it can all be in one merged pull?

monocongo:
Travis build is still failing, it seems like the issue is that nose is cooked into the numpy.testing package, which is being used in some of our tests. I'll look into whether or not this can be bypassed, and if not maybe getting away from numpy.testing, or better yet to just include nose as a dependency for tests only. I'm not sure how this is done in Python, but my assumption is that this is somehow possible (I'm thinking of how this is done in Maven for Java projects). 

Update: The first line of the introduction on this page confirms the necessity of nose: https://github.com/numpy/numpy/blob/master/doc/TESTS.rst.txt

kikocorreoso:
if #64 is accepted *.travis.yaml* and *environemnt.yaml* should be modified in order to do the same a little bit different but, IMHO, in a more standard way.

PR #64, opened 2018-01-25
New version of setup.py

kikocorreoso:
With this *setup.py* you can use 

`pip install .`

to install the package (and only the package and not `misc`, `tests`,...). 

It takes care of dependencies so `requirements.txt` and `environment.yaml` are less necessary or not needed at all.

To run the tests you can use:

`python setup.py test`

And it will take care the non-lib dependencies like `nose`.

I've added some trove classifiers in order to help people know more about the package.

If the PR is accepted some changes should be done to *.travis.yaml* file in order to run the installation and test with the new functionality.


monocongo (changes requested):
This looks good, thanks!

Two further details:
1) I expect that we may also be able to eliminate the pycurl dependency, since this was recently replaced with calls to urllib.request
2) The license is GPL2 in the setup section at top but GPL3 in the classifiers section below. For now we're using BSD3, unless there's good case for another license (I'm still figuring out the nuances of this and determining the best to use) 

Can we address these in this file before the PR approval?

monocongo:
Can we safely remove requirements.txt once the setup.py is in place? I will look into where the file is used by outside services such as Travis CI, if I remember correctly this file is in place only for this purpose.

The environment.yml file is in place primarily for use with conda environment configuration. If this can also go away and we can use a single setup script ("one ring to rule them all") then all the better. But it looks like this is mandatory if you want to use the conda environment approach I've offered in the README. But maybe there's a better way? I'm open/flexible.

monocongo (dismissed):

kikocorreoso:
Have a look to this: https://packaging.python.org/discussions/install-requires-vs-requirements/

I've ammended the requirements and the license (I don't know why I read GPL somewhere...).

monocongo (approved):
looks good, thanks!

PR #169, opened 2018-06-28
fix link

benjimin:
Previously 404'd at https://github.com/monocongo/climate_indices/blob/master/drought.gov


monocongo (approved):
Link works now. Thanks for the fix!

Issue #219, opened 2018-12-09
Conversion of tests for migration from unittest to pytest


monocongo:
Pytest appears to be a simpler/better testing framework, for example see [this article with a comparison](https://www.slant.co/versus/9148/9149/~unittest_vs_pytest) and [this reddit discussion](https://www.reddit.com/r/Python/comments/5uxh22/unittest_vs_pytest/) for details.

[unittest2pytest](https://github.com/pytest-dev/unittest2pytest) may make this a no-brainer.

monocongo:
Resolved by #221

monocongo:
We need to update the documentation to show `pytest` as the command to use for launching tests.

PR #221, opened 2018-12-10
Updated tests for migration from unittest to pytest

monocongo:
Modified test files for migration to [pytest](https://docs.pytest.org/en/latest/), with updates to the TravisCI configuration and removal of the previous fixtures file (replaced by `conftest.py`).


monocongo:
Resolves issue #219

Issue #220, opened 2018-12-09
Store test/fixture arrays in *.npy files


monocongo:
Currently we build numpy arrays for test fixtures via `conftest.py`, and this file is unnecessarily large and hard to read as a result. We should instead save the fixture data arrays in *.npy files, and read these into fixture objects using [`numpy.load`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.load.html).

monocongo:
Resolved by PR #222

PR #222, opened 2018-12-11
Reading fixture arrays from *.npy files

monocongo:
Values for numpy arrays used as text fixtures now kept in `tests/fixture/*.npy`.


monocongo:
Resolves #220

PR #263, opened 2019-02-10
New version of process_divisions.py without netcdf_utils

monocongo:
`netcdf_utils` module deleted


monocongo:
Merging this into the develop branch, with the assumption that we'll next merge a new version of `process_divisions.py` from a fork by @haysengithub.

Issue #270, opened 2019-02-23
It doesn't effectively handle daily datasets which are not 366 days per year


lincystar:
The conversion functions utils.transform_to_366day and utils.transform_to_gregorian were missed to be included in the recently updated versions of the process_grid.py and process_divisions.py scripts.

monocongo:
The calendar conversion was replaced, converting from Gregorian to 366-day for indices computation, then back to Gregorian for output. Updates have been merged into the develop branch.

monocongo:
@lincystar please try the code from the develop branch, and/or refer me to a link where I can download your daily precipitation dataset in order to validate it myself. Please list any related issues you have here so I can be aware and address. Thanks!

go1me:
I find a bug in this PR. if we use -index all  …… --periodicity monthly and dont in put a pet file, the software will calculate the pet,  there will be a error in _compute_write_index on line 584, because the pet have not the argument of periodicity

monocongo:
Thanks for the heads up @haysengithub!

The highlighted bug seems to be fixed in the latest update to the develop branch.

monocongo:
Resolved by #274

PR #274, opened 2019-02-25
Updates of process_grid.py script for units handling and daily values

monocongo:
Now ignoring precipitation units for SPI and PNP, and reintroduction of a 366-day calendar for daily processing.


monocongo:
Resolves #270

Issue #340, opened 2019-09-13
Missing log messages


Laura-Guillory:
**Describe the bug**
This software used to print the start/end/elapsed time (which I found useful for very long runs). During a previous refactoring of the code this stopped happening.

Looking into the cause, I saw that despite a logger being set up for level `INFO` in \_\_main\_\_.py, the program actually runs on level `WARN`. This may be because each file uses basicConfig() to set its level individually. However, [basicConfig()](https://docs.python.org/2.7/library/logging.html#logging.basicConfig) is designed to only be run once, and does nothing for all subsequent calls. Its first call would be in lmoments.py, setting the logger to `WARN` level for the entire program:

climate_indices/lmoments.py, lines 7-11
```python
# ------------------------------------------------------------------------------
# set up a basic, global _logger
logging.basicConfig(level=logging.WARN,
                    format='%(asctime)s %(levelname)s %(message)s',
                    datefmt='%Y-%m-%d  %H:%M:%S')
```

A potential way to get around this and have each file log to a different level is to call `logger.setLevel()` after getting the logger, instead of setting the level in a basicConfig() call.

**To Reproduce**
1. Run anything (such as: process_climate_indices --index pet --periodicity monthly --netcdf_temp /data/nclimgrid_lowres_tavg.nc --var_name_temp tavg --output_file_base results/nclimgrid_lowres --multiprocessing all_but_one)
2. Program finishes without printing anything to the terminal.

**Expected behavior**
The program previously printed INFO level logs revealing start/end/run time, etc.

Example:
2019-09-13  12:29:36 INFO Start time:    2019-09-13 12:29:36.525875
2019-09-13  12:29:36 INFO Computing PALMERS
2019-09-13  12:42:52 INFO End time:      2019-09-13 12:42:52.485691
2019-09-13  12:42:52 INFO Elapsed time:  0:13:15.959816

**Desktop (please complete the following information):**
- OS: Windows 10
- Version: Python 3.7.3

Laura-Guillory:
I have a suggested fix for this and will be submitting a pull request for it.

monocongo:
Thanks in advance for your help, @Laura-Guillory!

monocongo:
Resolved by PR #341

PR #341, opened 2019-09-13
Fix for logging levels incorrectly set to WARN.

Laura-Guillory:
Moved logging level setting outside of basicConfig() to allow for different logging levels for each file.

This is a suggested fix for Issue #340


monocongo (approved):
Thanks for making these changes, this is better.

Issue #106, opened 2018-04-23
Add a log-logistic distribution fitting


monocongo:
[It's been shown](http://spei.csic.es/home.html#p9) that a log-logistic distribution fitting is optimal for SPEI calculations. Let's add this to complement the gamma and Pearson III distribution fittings we already have available. This may already be available in some form in [statsmodels ](https://www.statsmodels.org/stable/index.html)or [SciPy](https://www.scipy.org/), we should investigate there before rolling our own.

monocongo:
Apparently, the log-logistic distribution is also referred to as the [Fisk distribution](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.fisk.html#scipy.stats.fisk), available within SciPy.

monocongo:
Related to issue #190.

Emmadd:
I created a draft (there are some minor issues) pull request for a Fisk implementation in #440
However, looking into the R code some more, I realised that the R SPEI package uses a generalised logistic distribution, more similar to [genlogistic](https://docs.scipy.org/doc/scipy/reference/reference/generated/scipy.stats.genlogistic.html#scipy.stats.genlogistic) in SciPy than to Fisk. The distributions might be similar when c=-/k, but I am not certain.

monocongo:
Thanks for your work on this @Emmadd 

Is it useful for us to also provide a generalized logistic distribution? I'd like to include it if so, but only if truly useful as it will add maintenance cost, and this project is already due significant maintenance, we don't want to add even more technical debt.

Emmadd:
I don't think so. But if someone with more statistical knowledge thinks otherwise it would be good to know.

Emmadd:
After some further investigation it seems the python SciPy Fisk option is the better one to use. The genlogistic in python is a special generalised logistic distribution, not THE generalised logistic distribution that R uses, which resembles Fisk. So #440 can be continued.

monocongo:
## Feasibility assessment (2026-09-22, against `main` b079dee7)

Short version: the plumbing is small now; the hard part is deciding **which** log-logistic, and picking wrong locks a public enum name onto the wrong distribution. Discussion opened at #1126 — please weigh in, especially if you have used these distributions in practice.

### The trap

This thread (and closed PR #440) assumes SciPy `fisk` is the distribution R `SPEI` fits. It is not:

[...]

PR #440, opened 2021-08-26
added log-logistic/fisk distribution to SPEI

Emmadd:
My first attempt to include the Fisk distribution into SPEI as suggested in #106, copying as mush as possible from the PearsonIII implementation. The last @numba.jit leads to problems, but code runs fine without. I have not created any tests or stuff yet. Hope someone can have a look :-) let me know what you think and what should be done.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -31,7 +31,7 @@ The following indices are provided:
 -  `SPI <https://climatedataguide.ucar.edu/climate-data/standardized-precipitation-index-spi>`__,
    Standardized Precipitation Index, utilizing both gamma and Pearson Type III distributions
 -  `SPEI <https://www.researchgate.net/publication/252361460_The_Standardized_Precipitation-Evapotranspiration_Index_SPEI_a_multiscalar_drought_index>`__,
-   Standardized Precipitation Evapotranspiration Index, utilizing both gamma and Pearson Type III distributions
+   Standardized Precipitation Evapotranspiration Index, utilizing both gamma, log-logistic (Fisk) and Pearson Type III distributions
 -  `PET <https://www.ncdc.noaa.gov/monitoring-references/dyk/potential-evapotranspiration>`__,
    Potential Evapotranspiration, utilizing either `Thornthwaite <http://dx.doi.org/10.2307/21073>`_
    or `Hargreaves <http://dx.doi.org/10.13031/2013.26773>`_ equations
@@ -345,11 +345,12 @@ SPEI monthly
 --calibration_end_year 2010 --multiprocessing all``
 
 The above command will compute SPEI (standardized precipitation evapotranspiration index,
-both gamma and Pearson Type III distributions) from input precipitation and potential evapotranspiration datasets
+both gamma, log-logistic (Fisk) and Pearson Type III distributions) from input precipitation and potential evapotranspiration datasets
 (in this case, the reduced resolution nClimGrid precipitation and PET datasets provided in the example inputs directory).
 The input datasets are monthly data and the calibration period used will be Jan. 1951 through Dec. 2010. The index
 datasets will be computed at 9-month and 18-month timescales. The output files will be
 `<out_dir>/nclimgrid_lowres_spei_gamma_09.nc`, `<out_dir>/nclimgrid_lowres_spei_gamma_18.nc`,
+`<out_dir>/nclimgrid_lowres_spei_fisk_09.nc`, `<out_dir>/nclimgrid_lowres_spei_fisk_18.nc`,
 `<out_dir>/nclimgrid_lowres_spei_pearson_09.nc`, and `<out_dir>/nclimgrid_lowres_spei_pearson_18.nc`.
 Parallelization will occur utilizing all CPUs.
 
--- a/src/climate_indices/compute.py
+++ b/src/climate_indices/compute.py
@@ -743,3 +743,292 @@ def transform_fitted_gamma(
     # as determined by the normal distribution's quantile (or inverse
     # cumulative distribution) function
     return scipy.stats.norm.ppf(probabilities)
+
+
+# ------------------------------------------------------------------------------
+@numba.jit
+def fisk_parameters(
+        values: np.ndarray,
+        data_start_year: int,
+        calibration_start_year: int,
+        calibration_end_year: int,
+        periodicity: Periodicity,
+) -> (np.ndarray, np.ndarray, np.ndarray, np.ndarray):
+    """
+    This function computes the probability of zero and fisk
+    distribution parameters corresponding to an array of values.
+
+    :param values: 2-D array of values, with each row representing a year
+        containing either 12 values corresponding to the calendar months of
+        that year, or 366 values corresponding to the days of the year
+        (with Feb. 29th being an average of the Feb. 28th and Mar. 1st values for
+        non-leap years) and assuming that the first value of the array is
+        January of the initial year for an input array of monthly values or
+        Jan. 1st of initial year for an input array daily values
+    :param periodicity: monthly or daily
+    :return: four 1-D array of fitting values for the fisk
+        distribution, with shape (12,) for monthly or (366,) for daily
+
+        returned array 1: probability of zero
+        returned array 2: first fisk distribution parameter (c)
+        returned array 3: second fisk distribution parameter (scale)
+        returned array 4: third fisk distribution parameter (loc)
+    """
+
+    # reshape precipitation values to (years, 12) for monthly,
+    # or to (years, 366) for daily
+    if periodicity is Periodicity.monthly:
+
+        values = utils.reshape_to_2d(values, 12)
+
+    elif periodicity is Periodicity.daily:
+
+        values = utils.reshape_to_2d(values, 366)
+
+    else:
+
+        raise ValueError("Invalid periodicity argument: %s" % periodicity)
+
+    # validate that the values array has shape: (years, 12) for monthly or (years, 366) for daily
+    if len(values.shape) != 2:
+        message = "Invalid shape of input data array: {shape}".format(shape=values.shape)
+        _logger.error(message)
+        raise ValueError(message)
+
+    else:
+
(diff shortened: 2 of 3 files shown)


Emmadd:
I was wondering as I was making changes to the docs whether they should not include the possibility of pre-computing SPEI  distribution fitting variables. This seems possible in the code to me.

monocongo:
> I was wondering as I was making changes to the docs whether they should not include the possibility of pre-computing SPEI distribution fitting variables. This seems possible in the code to me.

Yes, I agree, we have done this already for SPI, there is no reason we couldn't do something very similar for SPEI. Good idea!

Emmadd:
Hmmm, the tests that fail are for SPI and Pearson, are they related to things I've changed in the branch?
`=========================== short test summary info ============================
331
FAILED tests/test_compute.py::test_transform_fitted_pearson - AssertionError: 
332
FAILED tests/test_indices.py::test_spi - AssertionError: 
333
================== 2 failed, 34 passed, 11 warnings in 2.69s ===================`

monocongo:
Yes, almost certainly. This is why the tests are in place, to shield from introducing changes that will give different results from before. In this case, however, we're expecting to see different values since we're using a new distribution, so we should come up with an entirely new test result. For example, we expect certain results when using the Pearson fitting and other results when using gamma. We now need to come up with an expected result when we use this new fitting, so when the next person changes some code we'll have a test in place to make sure they haven't broken things for this distribution fitting. So for this new distribution, we need to have a separate collection of tests that exercise the new code and produce a result we can feel confident about. Once this is in place then all new code must pass the test, otherwise, we'll know we're introducing a significant/breaking change.

With all that being said I'm not certain that the code you've introduced here is actually breaking things. These checks went off the rails a year or so ago and I just haven't had time to do the requisite house cleaning to get us back on track. I expect what's up here is that you may have broken some things and others were already broken. I can run these tests locally to debug but it will take time, my day job is demanding. I totally appreciate your help with this package, don't forget that many (usually invisible) others will benefit!

monocongo:
No, @Emmadd I don't think your new code has caused any of the existing tests to fail. Several of them were broken before you started this work, and I have not found the time to maintain the package to get that back in order. Your new code looks legit (very nice work, thanks for all the comments!) and my only recommendation at this point is to add a few tests into the test suite for this new distribution fitting. For example, in tests/test_compute.py we have test_gamma_parameters and test_pearson_parameters so now we should add a test_fisk_parameters function, etc. Once those are in place and working as expected then I will approve this merge. Hopefully, I will soon find the motivational impetus to get the rest of our tests back in shape and this project back on the rails!

monocongo:
There was a fix merged today (#450) which may fix the failing tests here. I'm not sure how to get the checks to run again (the instructions in GitHub help pages didn't include the same buttons as I see in the web UI etc.), I think it can't happen since the original run of the checks happened over 30 days ago. Maybe you can pull from master to get the latest changes, then make a commit of that to this fork branch, which would then trigger a new check?

WeatherGod:
You can merge master back into your branch. Then push your branch back up. That will retrigger tests.

Turns out the science lead on my project is also interested in this particular distribution, so I'll let you know if he has any feedback. I should double-check the implementation here because it borrows from the pearson III code which I had to fix.

As for the option to save out the fitted params for SPEI, stay tuned... I have a draft branch with that feature for pearson3 and gamma.

WeatherGod:
Oh, before merging master back into your branch, you should fetch master from this repo. Let me know if you have questions!

src/climate_indices/compute.py
```diff
@@ -743,3 +743,267 @@ def transform_fitted_gamma(
...
+    :param loc: third fisk parameter, the location of the distribution
+    """
+
+    # only fit to the distribution if the values array is valid/not missing
+    if not np.all(np.isnan(values)):
+
+        # minimums_possible = _minimum_possible(c, loc, scale)
+        # minimums_mask = values <= minimums_possible
+        zero_mask = np.logical_and((values < 0.0005), (probabilities_of_zero > 0.0))
+        trace_mask = np.logical_and((values < 0.0005), (probabilities_of_zero <= 0.0))
```
WeatherGod, line 897: So, should these go away? Furthermore, if probabilities_of_zero isn't used, it should get taken out of the call signature here, and the `fisk_params()` should be updated to not return it.

src/climate_indices/compute.py
```diff
@@ -743,3 +743,267 @@ def transform_fitted_gamma(
...
+                        with size == # of years * 12
+                        'daily' indicates an array of full years of daily values
+                        with 366 days per year, as if each year were a leap year
+                        and any missing final months of the final year filled
+                        with NaN values, with array size == (# years * 366)
+    :param probabilities_of_zero: pre-computed probabilities of zero for each
+        month or day of the year
+    :param locs: pre-computed loc values for each month or day of the year
+    :param scales: pre-computed scale values for each month or day of the year
+    :param skews: pre-computed skew values for each month or day of the year
```
WeatherGod, line 960: No `skews` argument, and missing documentation for `c`. Also, I am not a fan of single letter variable names. Is this `c` parameter known by any other names?

src/climate_indices/indices.py
```diff
@@ -315,6 +316,35 @@ def spei(
...
             )
 
+    elif distribution is Distribution.fisk:
+
+        # get (optional) fitting parameters if provided
+        if fitting_params is not None:
+            probabilities_of_zero = fitting_params["probabilities_of_zero"]
+            locs = fitting_params["locs"]
+            scales = fitting_params["scales"]
+            c = fitting_params["c"]
```
WeatherGod, line 326: This is going to conflict a bit with #453.

docs/index.rst
```diff
@@ -31,7 +31,7 @@ The following indices are provided:
 -  `SPI <https://climatedataguide.ucar.edu/climate-data/standardized-precipitation-index-spi>`__,
    Standardized Precipitation Index, utilizing both gamma and Pearson Type III distributions
 -  `SPEI <https://www.researchgate.net/publication/252361460_The_Standardized_Precipitation-Evapotranspiration_Index_SPEI_a_multiscalar_drought_index>`__,
-   Standardized Precipitation Evapotranspiration Index, utilizing both gamma and Pearson Type III distributions
+   Standardized Precipitation Evapotranspiration Index, utilizing both gamma, log-logistic (Fisk) and Pearson Type III distributions
```
WeatherGod, line 34: missing Oxford comma here and below.

WeatherGod (changes requested):
Yeah, there is a lot of boilerplate here. Would be interesting to find out if a bunch of this can still get simplified.

Not sure, but I don't think any of these changes actually adds support in the main scripts, either. And I do think unit tests should get added. It is what caught the issues with Pearson III.

src/climate_indices/compute.py
```diff
@@ -743,3 +743,267 @@ def transform_fitted_gamma(
...
+            probability_of_zero = number_of_zeros / number_of_non_missing
+
+        # get the estimated parameters, if we have
+        # more than three non-missing/non-zero values
+        if (number_of_non_missing - number_of_zeros) > 3:
+
+            # get the fisk parameters for this time
+            # step's values within the calibration period
+            c[time_step_index], scales[time_step_index], locs[time_step_index], = \
+                scipy.stats.fisk.fit(time_step_values)
```
WeatherGod, line 864: Just had a thought about this last night. How good is this fit that uses Maximum Likelihood Estimation? For pearson3, the L-moments method was used (which needed to be implemented because scipy stats doesn't have it as an option), and for gamma, the Maximum Likelihood Estimation was used, but via the Thom method, which also isn't available through scipy.stats. I have no experience in judging various parameter estimations, but we probably should have some sort of justification for which approach is used. I think having an existing, well-cited paper would be a good standard.

WeatherGod:
This PR also got me thinking about my work-in-progress adding SPEI support to the "spi" script. Right now, the spi script pre-allocates fitting arrays and stores them in a dictionary keyed by their parameter name. With the two current distributions, the fitting parameter names do not collide with each other. However, it seems like this distribution would have colliding parameter names. I'll think on this a bit, but an elegant solution may require a significant overhaul of the "spi" script, and possibly of the other main script, too (which isn't a bad thing!).

In addition, we should also consider updating the API and CLI so that one can select which distributions to compute for in order to keep computational and memory costs down.

src/climate_indices/compute.py
monocongo: Agreed, meaningful variable names are preferred.

src/climate_indices/indices.py
monocongo: Yes, we should/will normalize.

src/climate_indices/compute.py
monocongo: I originally wanted to add this since the Spanish authors who are most famous for SPEI used this distribution for much of their work. They have provided [an implementation in R](https://cran.r-project.org/web/packages/SPEI/SPEI.pdf). The methodology is nicely described [here](https://www.researchgate.net/publication/291185005_SPEI-Based_Spatiotemporal_Analysis_of_Drought_in_Haihe_River_Basin_from_1961_to_2010) (page 3)

Here is an interesting [goodness of fit analysis](https://rstudio-pubs-static.s3.amazonaws.com/202764_0f164cddb7fd415eae1389ea27cec447.html)

WeatherGod: I love papers like this! Very clearly lays out the algorithm, both in words and in equations, and also provides names for some of the steps they use. It is clear from this paper that an L-moments method is used (but seems different from the L-moments used for pearson III). The formulation is also different enough to give me pause, but that just might be that the scipy documentation is showing a standardized form while the paper isn't. It is going to take a bit of algebra to cross-check the paper's L-moments with the parameters used in scipy's implementation.

On a science note, I am a bit confused by the "non-negative" aspect of this distribution, given that Pt - PET can be negative, but I guess that is what the "loc" (in scipy) or "gamma" (in paper) can account for?

Issue #419, opened 2021-03-14
tox reports 2 failed tests for py37


cshields143:
**Describe the bug**
`test_compute.py::test_transform_fitted_pearson` throws an AssertionError:

```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0.001
E       Transformed Pearson Type III fitted values not computed as expected
E       x and y -inf location mismatch:
E        x: array([[ 0.207576,  0.891472, -1.00923 , ...,  0.447044,  0.735171,
E               -0.072223],
E              [ 0.724334,  0.393734, -0.988155, ..., -0.266911,  0.684934,...
E        y: array([[ 0.207576,  0.891472, -1.00923 , ...,  0.447044,  0.735171,
E               -0.072223],
E              [ 0.724334,  0.393734, -0.988155, ..., -0.266911,  0.684934,...
```

Ditto for `test_indices.py::test_spi`:

```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0.01
E       SPI/Pearson values for 6-month scale not computed as expected
E       Mismatched elements: 120 / 1476 (8.13%)
E       Max absolute difference: 6.18
E       Max relative difference: 2.
E        x: array([nan, nan, nan, ..., nan, nan, nan])
E        y: array([nan, nan, nan, ..., nan, nan, nan])
```

**To Reproduce**
1. `git clone https://github.com/monocongo/climate_indices`
2. `cd climate_indices`
3. `conda create -n python37 python=3.7`
4. `conda activate python37`
5. `pip install -r requirements.txt`
6. `pip install tox`
7. `tox -e py37`

**Expected behavior**
`py37: commands succeeded`
`congratulations :)`

**Desktop (please complete the following information):**
- OS: Ubuntu
- Version: 20.04.2 LTS
- Architecture: 64bit

**Additional context**
All tests *do* pass with python 3.6

cshields143:
Obviously I have the source, so I can do some debugging and see if I can't figure out why.

Maybe I don't have the environment or testing pipeline set up right?

cshields143:
Some "progress" has been made on the first failed test: `scipy.stats.pearson3.cdf` returns different values when run on different python versions. For more info: https://github.com/cshields143/climate_indices_issue_419a

It returns `0` in later versions (3.7, 3.8) when it returns `nan` in earlier versions (3.6)

monocongo:
[...]

PR #448, opened 2021-10-21
Can't assume that the input's coordinates are in the same order as the output's dimensions

### LINKED ISSUE #445: Getting error: conflicting sizes for dimension 'lat' for SPI calculation from NetCDF
**[How to craft a useful, minimal bug report](http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports)**

**Describe the bug**
I'm trying to calculate SPI from Chirps data in NetCDF format and I'm getting 

"ValueError: conflicting sizes for dimension 'lat': length 12 on the data but length 170 on coordinate 'lat'" error from XArray. I've been using the 1.0.9 version, but also tried the 1.0.8 and 1.0.7 versions with the same error. 

Any ideas to solve the issue?

**To Reproduce**
I'm trying to create SPI with the following command :

spi_chirps_2021_v2 % spi --periodicity monthly --scales 12 --calibration_start_year 1981 --calibration_end_year 2020 --netcdf_precip /_temp/spi_chirps_2021_v2/input_nc/tr_clip_chirps_1months_1980_2021.nc --var_name_precip precip --output_file_base /_temp/spi_chirps_2021_v2/output_nc/tr_chirps --multiprocessing all --save_params /_temp/spi_chirps_2021_v2/Fitting/tr_chirps_fitting.nc --overwrite

**Desktop (please complete the following information):**
- OS: MacOS X 10.15.7
- Python Version : 3.7.11

WeatherGod:
* Closes #445


WeatherGod:
Note, I think this bug also is sort-of present in `__main__.py`, but it is hard to be sure. It makes some slightly different assumptions and it might all be ok when it comes out in the wash. I have generated SPI values using this patch on my own datasets through the `spi` command, and I have generated SPEI values using the same datasets through the `process_climate_indices` command without issues that I have noticed so far.

WeatherGod:
Failures noted by DeepSource are unrelated to this PR.

monocongo:
The failing tests are not caused by this PR. The tests went off the rails a while back and I've not had time to right the ship, another user has identified the possible cause, see #419.

monocongo:
What's the best way forward? My thinking is that it's putting the cart before the horse to approve PRs with failing tests, but fixing the tests might be a long/laborious process (one reason I've procrastinated) and I don't want to hold up good PRs on account of it.

WeatherGod:
Well, I have some good news for you. I just took a look at one of the test failures, and I figured out what is happening and I think I know the correct thing to do. I'll submit the PR for that and let you mull it over as you have better historical context for the relevant chunk of code (looks like the code got moved around, so the history is difficult to figure out). I'll look at the other one next and see if I can figure it out, too.

monocongo:
That is such good news and totally appreciated. Once the tests are working as expected again then reviewing/approving all of the outstanding PRs will be easier.

WeatherGod:
Rebased and force-pushed to include the recent fixes.

PR #450, opened 2021-10-22
Properly mask out estimated below minimum CDF vals in the Pearson calcs

WeatherGod:
* Fixes a test failure where values that were below the estimated minimum
  were becoming -inf instead of being set to trace.
* Don't see a reason why a nan mask was getting involved, they only
  happen where the original data had NaNs, and you'd want the fitted
  values to remain NaN there, right?

### DIFF
--- a/src/climate_indices/compute.py
+++ b/src/climate_indices/compute.py
@@ -412,8 +412,7 @@ def _pearson_fit(
         # that threshold then we set the corresponding CDF to a floor value
         # TODO ask Richard Heim why the use of this floor value, matching
         #  that used for the trace amount?
-        nans_mask = np.isnan(values)
-        values[np.logical_and(minimums_mask, nans_mask)] = 0.0005
+        values[minimums_mask] = 0.0005
 
         # account for negative skew
         skew_mask = skew < 0.0
--- a/src/climate_indices/compute.py
+++ b/src/climate_indices/compute.py
@@ -1,4 +1,5 @@
 from enum import Enum
+from distutils.version import LooseVersion
 import logging
 
 # from dask.array import pad
@@ -7,6 +8,9 @@ import numba
 import numpy as np
 import scipy.special
 import scipy.stats
+import scipy.version
+
+_do_pearson3_workaround = LooseVersion(scipy.version.version) < '1.6.0'
 
 from climate_indices import utils, lmoments
 
@@ -394,9 +398,16 @@ def _pearson_fit(
 
     # only fit to the distribution if the values array is valid/not missing
     if not np.all(np.isnan(values)):
-
+        # This is a misnomer of sorts. For positively skewed Pearson Type III
+        # distributions, there is a hard lower limit. For negatively skewed
+        # distributions, the limit is on the upper end.
         minimums_possible = _minimum_possible(skew, loc, scale)
-        minimums_mask = values <= minimums_possible
+        minimums_mask = (values <= minimums_possible) & (skew >= 0)
+        maximums_mask = (values >= minimums_possible) & (skew < 0)
+
+        # Not sure what the logic is here given that the inputs aren't
+        # standardized values and Pearson III distributions could handle
+        # these sorts of values just fine given the proper parameters.
         zero_mask = np.logical_and((values < 0.0005), (probabilities_of_zero > 0.0))
         trace_mask = np.logical_and((values < 0.0005), (probabilities_of_zero <= 0.0))
 
@@ -408,15 +419,30 @@ def _pearson_fit(
         values[zero_mask] = 0.0
         values[trace_mask] = 0.0005
 
-        # compute the minimum value possible, and if any values are below
-        # that threshold then we set the corresponding CDF to a floor value
-        # TODO ask Richard Heim why the use of this floor value, matching
-        #  that used for the trace amount?
-        values[minimums_mask] = 0.0005
+        if _do_pearson3_workaround:
+            # Before scipy 1.6.0, there were a few bugs in pearson3.
+            # Looks like https://github.com/scipy/scipy/pull/12640 fixed them.
 
-        # account for negative skew
-        skew_mask = skew < 0.0
-        values[:, skew_mask] = 1 - values[:, skew_mask]
+            # compute the minimum value possible, and if any values are below
+            # that threshold then we set the corresponding CDF to a floor value.
+            # This was not properly done in older scipy releases.
+            # TODO ask Richard Heim why the use of this floor value, matching
+            #  that used for the trace amount?
+            nans_mask = np.isnan(values)
+            values[np.logical_and(minimums_mask, nans_mask)] = 0.0005
+            # This will get turned into 0.9995 when the negative
+            # skew bug is worked around a few lines from here.
+            values[np.logical_and(maximums_mask, nans_mask)] = 0.0005
+
+            # account for negative skew
+            skew_mask = skew < 0.0
+            values[:, skew_mask] = 1 - values[:, skew_mask]
+        else:
+            # The original values were found to be outside the
+            # range of the fitted distribution, so we will set
+            # the probabilities to something just within the range.
+            values[minimums_mask] = 0.0005
+            values[maximums_mask] = 0.9995
 
         if not np.all(np.isnan(values)):
 


WeatherGod:
As for the other test error, I see an interesting pattern. All of the mismatched values computed for pearson3-fitted SPI happen on September for the entire timespan. Weird...

WeatherGod:
(and that happens with or without this patch...)

WeatherGod:
The signs are flipped!!!!

```
(Pdb) computed_spi[-10:, 8]
array([ 0.67895778,  0.1463942 , -0.68888282, -0.13619928, -1.92770816,
       -0.68096555,  1.35666739, -0.2893708 ,  0.93860201,         nan])
(Pdb) spi_6_month_pearson3[-10:, 8]
array([-0.67895778, -0.1463942 ,  0.68888282,  0.13619928,  1.92770816,
        0.68096555, -1.35666739,  0.2893708 , -0.93860201,         nan])
```

Now that is really weird!

monocongo:
There's a single test failing now for SPI, we're almost there! This is excellent progress...

WeatherGod:
The fitted skew for september is coming out negative, unlike the rest of the months. That could be causing signs to flip, maybe?

src/climate_indices/compute.py
```diff
@@ -412,8 +412,7 @@ def _pearson_fit(
         # that threshold then we set the corresponding CDF to a floor value
         # TODO ask Richard Heim why the use of this floor value, matching
         #  that used for the trace amount?
-        nans_mask = np.isnan(values)
-        values[np.logical_and(minimums_mask, nans_mask)] = 0.0005
+        values[minimums_mask] = 0.0005
 
         # account for negative skew
         skew_mask = skew < 0.0
```
WeatherGod: So, if I don't do this negative skew mask, the other test will pass. I have no clue if the code is incorrect or if the test data is incorrect. Any ideas where this stuff came from? I wonder if perhaps we could do a round-trip test (i.e., data -> fitted -> data) with the last step being done by something in scipy? It has been 15 years since my last stats class, and I certainly never played around with Pearson distributions.

WeatherGod: I have searched high and low, and while I have found and verified the python vs. original fortran implementations of the lmoments code, I just can't find any authoritative reference for this `_pearson_fit()` implementation. And my understanding of the stats math is poor, so I don't know what is right and what is wrong.

monocongo: It was maybe ten years ago when I wrote some version of this in Python which was a naive translation of the original Fortran. It has somehow evolved into what we see here. I also don't fully understand the statistics part of this, but it seems to work well for the folks who do. As we can see in the comments above (line 413), there are implementation details that never quite made sense to me. 

As far as where the original test data came from: I cooked it up from nClimGrid or nClimDiv datasets once the implementation's results closely matched those we'd get from the previous Fortran codes. That's how it worked for SPI, and for SPEI we probably did something similar but maybe using the Spanish SPEI folks' maps or WRCC DRI maps as references to compare against?

WeatherGod: Just had an inspiration about where we can find more authoritative information. I went back to the GPL'ed library I was originally looking at to see what sort of references they were using without looking at their code. This PDF looks promising: https://github.com/e-baumer/standard_precip/blob/master/docs/lloyd_hughes.pdf

WeatherGod: Also, on Monday, I'll check in with our project's head scientist to see what references he has for using Pearson type III distributions.

monocongo: Yes, that paper indicates that their implementation of the gamma distribution fitting worked best for Europe. [Guttman (1998)](https://drought.unl.edu/archive/Documents/NDMC/Workshops/136/Ref/SPI_Guttman.pdf) votes in favor of the Pearson III for the USA, not because it's any better than gamma but because it has more parameters:  
```
Because the PE3 with its three parameters is more flexible than the GAM with its two parameters in fitting sample data, it is suggested that the PE3 be used as a universal model from which to compute the SPI from site data. Although not discussed, the PE3 model is also the best choice for describing wet events.
```
Since Guttman worked in the same office they went with his suggestion, and I eventually inherited the code when I was hired years later. There were numerous versions in play at NOAA and several others in the wild, each one computing things slightly differently with varying measures of agreement across the spectrum, and I thought this shouldn't be, there should instead be a community reference implementation that everyone can converge on as the right way to compute SPI (and the other indices introduced later), then there can be less re-invention of the wheel and more standardization of calculations. With a commonly used open-source package then studies have a much better chance at being reproducible, unlike the two papers we've referenced above -- I'll bet dollars to doughnuts it can't be done for either one since the code is lost to history. But I digress...

WeatherGod: A very laudable goal, and putting it together into a package like this is a good first step. Having unit tests is also important. The question here is... which one is "correct" - the test data or the implementation? One or the other is going to have to change in order to fix this test (I'd hate to throw out the test).

I still haven't found any reference to what seems like a survival function right in the middle of this Pearson normalization. From what I can tell, negative skews are "a bad thing" in gamma distributions, but I still don't see anything claiming that this is how to address a negative skew or even that it does need to be addressed at all. Although, the absence of evidence isn't evidence of absence. This isn't my field (I'm a big data / remote sensing guy), so I don't even know where to start.

At the moment, I'm inclined to add a commit that takes this out and preserves the test data.

monocongo: The test data is just the best effort at getting what we think are good values and using those going forward to make sure we haven't gone off the rails. We just decide at some point that the numbers  look good and our implementation is solid and then go with it until someone can show where/how we're off. It's a chicken and egg problem.

Something I came across over the weekend was a line where the Pearson parameter (skew) value gets inverted if an L-moment is out-of-range, [here](https://github.com/monocongo/climate_indices/blob/master/src/climate_indices/lmoments.py#L101):
```
        if lmoments[2] < 0:
            skew = -2.0 / alpha_root
        else:
            skew = 2.0 / alpha_root
```
Maybe this has something to do with it?

WeatherGod: I saw that too, but that is actually correct. It is present in the original fortran code and I see it done in some other implementations in R and such. I think that is solid.

The lead scientist for our project just pointed me to this publication that does show a conditional handling of negative skewness, however the formulation is different (it is Pearson3 as implemented through a gamma function): http://www.tc.faa.gov/its/worldpac/techrpt/ar03-62.pdf. In particular, equations 4 and 5.

monocongo: This is an excellent find, thank you.

It would be interesting to implement this approach and then compare it against the L-moments-based implementation we have in place now. If we can reproduce the numbers provided in this paper then it'd be nice to be able to match with such solid research.

WeatherGod: I am trying to wade myself through this. Unfortunately, this document seems to use a different formulation of the gamma distribution than is used by scipy.stats, and seems to take an assumption that beta = 1 for some parts, but that doesn't seem to make any sense to me. I am getting myself lost, here.

WeatherGod: Made a little bit of progress this morning with understanding the math, but I just keep going around in circles. Going to put this aside for now until I get the chance to grab my old stats book from my office this week.

monocongo: It's turtles all the way down.  /s

This is exactly the type of in-depth vetting I was hoping for when I open-sourced this package. Godspeed!

WeatherGod: Learning lots of good stuff from my "Statistical Methods in the Atmospheric Sciences" book, or at least, good clarifying information. Still haven't completely cracked this nut, but one possible theory I am mulling in my head is that the negative skew handling may have been a misguided attempt at making the PDF artificially handle negative skews by making it symmetric w.r.t. skew. I have found another application of gamma distributions where something like this is done, but not for skew. The pearson III distribution already allows for negative skew, so it just makes me think the code is incorrect. But I am still not sure.

WeatherGod: Alright, I have been doing some plotting, and without knowing exactly the rationale _for_ doing the `1 - values` when the skew is negative, I am now convinced that I could never come up with my own justification for doing so. And just to show, for the test data that is causing the failure, here is its histogram and its fitted pearson3 distribution.
image

And here is what the transformed values look like with and without the "negative skew modification". Note that the blue points (no modification for negative skew) look a lot like the transformed values for the months with positive skews.
image

WeatherGod: Ah! I think I found a reason! https://github.com/scipy/scipy/issues/11186

There was a bug in earlier versions of scipy that caused the cdf to be incorrect for negative skews! This was only fixed in the past year, it looks like!

WeatherGod: Also looks like the same set of fixes also fixed an issue that might have involved the nan mask, too. Maybe this means we need a scipy version check to decide what to do.

WeatherGod: Let this be a lesson: when working around a bug in another library, file a bug report and include a link to that bug report in the comments for the work-around.

WeatherGod:
Alright, I am working out a final version of the fix. Turns out there is some sort of nuance I missed with the minimums masking that I am still sorting out. My original fix didn't actually solve all situations with newer scipy's. Seems like one test wants the fix, while another test doesn't want the fix. Weird...

WeatherGod:
Done! So, not only did we need to version-gate the work-arounds, but also there was a pre-existing lurking bug for negatively skewed data that was accidentally papered-over by the work-around. The handling of the minimum possible values is only valid for positively skewed data. You need a similar check for negatively skewed data since the limit is on the other end of the number line.

monocongo:
This is totally a "next level" effort, you went deep to work this out @WeatherGod. I truly appreciate this contribution, bravo!

src/climate_indices/compute.py
monocongo: Agreed, that's the way. For what it's worth we never knew there was a bug in scipy causing this work-around to be baked into our code, so when it was fixed we were caught with our pants down.

WeatherGod: Not completely. Your unit test did flag something, which is better than nothing. If you didn't have that one timeseries in the test data that just so happened to skew negatively, then your pants would have been down...

PR #451, opened 2021-10-25
[WIP] Have a FittingParams base class instead of inconsistently keyed dicti…

WeatherGod:
…onaries

### DIFF
--- a/src/climate_indices/fittings.py
+++ b/src/climate_indices/fittings.py
@@ -1,0 +1,118 @@
+from dataclasses import dataclass
+
+from climate_indices import compute
+
+
+class FittingParams:
+    @classmethod
+    def from_dict(klass, params):
+        """
+        Compatibility shim. Convert old accepted parameter dictionaries
+        and return specific instances of FittingParams subclasses.
+        Don't call this method from FittingParams, call it from the subclass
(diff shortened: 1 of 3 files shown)


WeatherGod:
Note, this particular implementation breaks support for py36.

Also, I am not 100% enthused by this exact organization. I kind of want there to be a Distribution base class, with instances that would implement their respective transformations and they would take a FittingParams instance that matched their distribution, or maybe what I really want is to call FittingParams Distribution instead? Feedback welcomed.

WeatherGod:
Sleeping on this a bit, I am starting to think that perhaps I should just rework `from_dict()` into a dict-normalizing private function. Really scale back this whole idea because I don't want to paint this project into a corner where it becomes stuck with these scalar-holding data structures. An idea that I have been toying with in my mind is to fully embrace xarray Datasets and putting fitting parameters in there alongside the input data. This data structure impedes that concept.

I am liking the idea, though, of using distribution as a way to pass down an object with some class methods to perform parameter fitting and data transformation. As you can see in `compute()`, it greatly simplifies the logic. I don't know how that might interfere with numba, though.

WeatherGod:
Closing in favor of #453.

PR #453, opened 2021-10-26
Normalize fitting keys between spi() and spei(). Closes #449.

### LINKED ISSUE #449: indices.spi and indices.spei expect differently named fitting params
`indices.spi()` and `indices.spei()` takes a dictionary of fitting params with differently named entries:

| SPI | SPEI |
| --- | ---- |
| alpha | alphas |
| beta | betas |
| prob_zero | probabilities_of_zero |
| loc | locs |
| scale | scales |
| skew | skews |

This makes it significantly more difficult to use `spi` and `spei` somewhat interchangeably (setting aside the fact that the latter takes two input variables...). I can normalize this, but it would technically be an API break. I was sort of wondering if instead we introduce a base class `FittingParams`, and have subclasses for each kind of distribution and pass that around. Then the functions can still accept dictionaries for backwards compatibility and also the FittingParams.

WeatherGod:
* Deprecate fitting parameter keys used in spei() that are
  inconsistent with the ones used in spi()

### DIFF
--- a/src/climate_indices/indices.py
+++ b/src/climate_indices/indices.py
@@ -31,6 +31,34 @@ _FITTED_INDEX_VALID_MIN = -3.09
 _FITTED_INDEX_VALID_MAX = 3.09
 
 
+def _norm_fitdict(params: Dict):
+    """
+    Compatibility shim. Convert old accepted parameter dictionaries
+    into new, consistently keyed parameter dictionaries. If given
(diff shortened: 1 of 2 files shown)


WeatherGod:
Supersedes #451.

WeatherGod:
Rebased and force-pushed to include recent fixes.

PR #467, opened 2022-03-21
comparing longitudes and latitudes with tolerance

### LINKED ISSUE #465: Consider a small tolerance when verifying matching coordinates
**Is your feature request related to a problem? Please describe.**
SPEI, for instance, requires two NetCDFs. If the float lon/lat of the same data provider differ slightly due to (presumably) rounding errors, `np.array_equal()` returns `False` which causes a `ValueError: Precipitation and temperature variables contain non-matching latitudes` in `_validate_args`.

**Describe the solution you'd like**
A small tolerance should be considered in the comparison of x1 and x2. A solution might be `np.allclose(x1, x2, rtol=0, atol=small_threshold)`. In my case (grid resolution of 0.1°), `small_threshold` of 1e-5 would be fine. This is already done in other comparisons (e.g. lats_precip and dataset_awc). I suggest to consequently replace the `np.array_equal`s with `np.allclose`s.

**Describe alternatives you've considered**
`np.abs(x1-x2).max() < small_threshold`

**Additional context**
I issue occurred with [E-OBS](https://www.ecad.eu/download/ensembles/download.php) datasets, i.e. temperature and precipitation.

itati01:
This is to close #465. The remaining `np.array_equal`s are replaced by `np.allclose`s where longitudes and latitudes are compared but no divisions (`input_type == InputType.divisions`). The fixed absolute tolerance of 0.01 is replaced by a dynamic tolerance derived from the (smallest) grid size.

### DIFF
--- a/src/climate_indices/__main__.py
+++ b/src/climate_indices/__main__.py
@@ -15,7 +15,8 @@ from climate_indices import compute, indices, utils
 
 # the number of worker processes we'll use for process pools
 _NUMBER_OF_WORKER_PROCESSES = multiprocessing.cpu_count() - 1
-
+# threshold absolute tolerance parameter np.allclose
+ATOL_THRESH = 0.01
 # shared memory array dictionary keys
 _KEY_ARRAY = "array"
 _KEY_SHAPE = "shape"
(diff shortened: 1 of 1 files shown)


monocongo:
This is a really nice update. My apologies for letting this languish. Kudos @itati01 !

src/climate_indices/__main__.py
```diff
@@ -90,6 +89,11 @@ def _validate_args(args):
                                          ("division", "time"),
                                          ("division")]
 
+    # dynamic threshold absolute tolerance parameter np.allclose
+    # derived from (smallest) grid size along dimension dim
+    def get_tolerance(dim):
+        return np.diff(dim).min() / 10
```
WeatherGod, line 95: This introduced a bug. If the latitudes are North to South, you get a negative absolute tolerance, and the `np.allclose()` will always fail.

WeatherGod: PR coming to fix this.

PR #477, opened 2022-09-26
Add a --chunksize option with "inputs" as a choice

### LINKED ISSUE #476: option to preserve chunking?
**Is your feature request related to a problem? Please describe.**
I am using process_climate_indices in a data pipeline, with this step being done on a fairly large, dedicated compute resource. The steps beforehand take advantage of some carefully crafted netcdf4 chunksizes, and this indirectly helps process_climate_indices with its multiprocessing.

There are some steps afterwards that operate on the netcdfs generated by process_climate_indices, but they aren't running as efficiently as I'd like because the outputs get saved out contiguously. Since compute_climate_indices ends up with the climate index entirely in memory, it can write it out in a chunked or contiguous manner with no real difference in performance.

**Describe the solution you'd like**
I'm thinking it would be a useful feature to utilize the chunksizes of the input files (if any) and apply them for the output files. Correct me if I'm wrong, but the input dimensions are always the same as the output dimensions, so we should be able to have a `--chunksizes inputs` option that would mean "use the input file's chunksizes in the output files". This could also lead to future abilities to explicitly specify chunksizes.

If this would be welcomed, I think I can put together a PR that does this in fairly short order.

WeatherGod:
* Enabling this allows one to carry the input file's chunksizes through to the output file(s).
* Closes #476

### DIFF
--- a/src/climate_indices/__main__.py
+++ b/src/climate_indices/__main__.py
@@ -709,6 +709,8 @@ def _compute_write_index(keyword_arguments):
         files.append(keyword_arguments["netcdf_pet"])
     if "input_type" not in keyword_arguments:
         raise ValueError("Missing the 'input_type' keyword argument")
+    if "chunksizes" not in keyword_arguments:
+        raise ValueError("Missing the 'chunksizes' keyword argument")
     else:
         input_type = keyword_arguments["input_type"]
         if input_type == InputType.grid:
@@ -719,7 +721,19 @@ def _compute_write_index(keyword_arguments):
             chunks = {"time": -1}
         else:
             raise ValueError(f"Invalid 'input_type' keyword argument: {input_type}")
-    dataset = xr.open_mfdataset(files, chunks=chunks)
+    # Since multiple variables can be in the same file, de-duplicate the filelist.
+    dataset = xr.open_mfdataset(list(set(files)), chunks=chunks)
+    output_chunksizes = {}
+    if keyword_arguments['chunksizes'] == 'input':
+        # Find the first variable with chunksizes set and use that
+        # Note that the netcdf spec doesn't require that all data variables
+        # have the same chunk sizes.
+        for da in dataset.data_vars.values():
+            if not da.encoding.get("contiguous", True):
+                # tuple of chunksizes, respectively by dimension
+                output_chunksizes = da.encoding.get("chunksizes", ())
+            if output_chunksizes:
+                break
 
     # trim out all data variables from the dataset except the ones we'll need
     input_var_names = []
@@ -802,6 +816,9 @@ def _compute_write_index(keyword_arguments):
 
     # build an arguments dictionary appropriate to the index we'll compute
     args = _build_arguments(keyword_arguments)
+
+    output_encodings = ({'chunksizes': output_chunksizes}
+                        if output_chunksizes else None)
 
     # add output variable arrays into the shared memory arrays dictionary
     if keyword_arguments["index"] == "palmers":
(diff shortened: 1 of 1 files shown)


WeatherGod:
Don't know if we want to give the same treatment to `__spi__.py` or not. Also, I wonder if we might be painting ourselves into a corner with the command-line option name. What if in the future, we would want the ability to chunk the processing itself from start to finish? Chunksizes for I/O tends to be different from chunksizes for processing.

WeatherGod:
Furthermore, do we want to have a similar ability to carry forward other encoding information like compression and such?

monocongo:
> Furthermore, do we want to have a similar ability to carry forward other encoding information like compression and such?

So we could have a `--preserve-encoding` flag which would trigger a snapshot of the data's configuration on the way in, we can then mangle the data as we see fit while processing the indices, then assemble the output in a way that preserves the encoding settings? This almost seems like it should be the default behavior, especially since my plan is to eventually remove the NCO code which makes an attempt to optimize the data before processing. That is the most problematic part of this code (well over 50% of the user help requests are a result of NCO issues, especially on Windows), and we can instead take a less "batteries included" approach and expect users to know how they want their data organized, and do the re-orientation/optimization internally with numpy rather than on the NetCDF itself, though it might require a bit more bookkeeping.

WeatherGod:
Indeed. Now that xarray and dask has significantly matured, a lot of the code here can eventually be removed, along with many of its restrictions. I will note that there is an advantage to having an explicit step that specifically reorganizes the data in a manner that is optimal for time-series analysis. The `rechunker` project (https://rechunker.readthedocs.io/en/latest/) talks about this a fair bit, but unfortunately, it is geared towards ZArr format rather than netcdf.

I'm thinking that `--preserve-encoding` would be the general, all-around way to preserve such information, while `--chunksizes inputs` would be the way to do just the chunksizes (or even, in the future, change the chunking on-the-fly). Other arguments could introduce compression options, again, overriding the compression-specific encodings from a `--preserve-encoding` flag.

src/climate_indices/__main__.py
```diff
@@ -719,7 +721,22 @@ def _compute_write_index(keyword_arguments):
...
+        # have the same chunk sizes.
+        for da in dataset.data_vars.values():
+            if not da.encoding.get("contiguous", True):
+                # tuple of chunksizes, respectively by dimension
+                output_chunksizes = da.encoding.get("chunksizes", ())
+            if output_chunksizes:
+                break
+
+    # Now apply the requested chunks
+    dataset = dataset.chunk(chunks)
```
WeatherGod: Actually, this isn't needed since the file(s) are loaded with the specified chunks.

WeatherGod:
Here is what the encoding looks like for the PET variable in one of my files as a reference:
```python
{'zlib': False, 'shuffle': False, 'complevel': 0, 'fletcher32': False, 'contiguous': False, 'chunksizes': (103, 240, 12), 'source': '.(redacted)/pet.nc', 'original_shape': (721, 1440, 872), 'dtype': dtype('float32'), '_FillValue': nan}
```
Some of this stuff shouldn't get "preserved" such as the `'source'`. `'dtype'` is a bit questionable.

monocongo:
Other than two test cases that are now somehow off I can approve this. The differences are small/minor from what I can see so far, probably if we compare the new results against the old we won't see anything obvious, but I've not gone that far with it yet. I had trouble getting HDF5/NetCDF etc. installed on the machine where I ran the tests, so I'm still not sure if anything like that may be off causing the minor discrepancies I've seen so far in the pytest output, but the overall approach here for preserving the chunk sizes looks solid.

WeatherGod:
Yeah, you're right, I shouldn't let perfect be the enemy of good enough. Is there anyplace else you want this documented? Like a "what's new" or somesuch?

monocongo:
I will work on a release to PyPI with a new version and what's new, so if you want to add some notes to that it'll be appreciated, @WeatherGod then stay tuned for the PR.

 Thanks again for your effort on this, very appreciated!

WeatherGod:
Sure, I can add some notes

monocongo:
BTW I pushed a new version to PyPI already, might have jumped the gun

PR #497, opened 2023-02-08
Force get_tolerance() to always be greater than zero.

WeatherGod:
* Also move it from __main__.py to utils.py and add some tests

### DIFF
--- a/src/climate_indices/utils.py
+++ b/src/climate_indices/utils.py
@@ -489,4 +489,15 @@ def count_zeros_and_non_missings(
     return zeros, non_missings
 
 
+# ------------------------------------------------------------------------------
+def get_tolerance(dim):
+    """
+    dynamic threshold absolute tolerance parameter np.allclose
+    derived from (smallest) absolute grid size along dimension dim.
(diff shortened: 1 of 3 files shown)


monocongo (commented):
The only "low-hanging fruit" changes I see are some type hints, for example on the `utils.get_tolerance` function's `dim` argument. Otherwise, this looks solid.

src/climate_indices/utils.py
```diff
@@ -489,4 +489,15 @@ def count_zeros_and_non_missings(
...
 
 
+# ------------------------------------------------------------------------------
+def get_tolerance(dim: np.ndarray) -> float:
+    """
+    dynamic threshold absolute tolerance parameter np.allclose
+    derived from (smallest) absolute grid size along dimension dim.
+    Always greater than zero.
+    """
+    tol = np.abs(np.diff(dim).min() / 10)
```
WeatherGod, line 499: Just realized that this is technically incorrect. It should be the minimum of the absolute value, not the absolute value of the minimum.

PR #499, opened 2023-02-09
order of operations fix

WeatherGod:
minimum of absolute values instead of absolute value of minimum


monocongo:
Thanks @WeatherGod !

PR #515, opened 2023-05-25
Update PDSI_Canadian_Stations.ipynb

Daafip:
pd.datetime didn't work for me, should be pd.Timestamp I think

### DIFF
--- a/notebooks/PDSI_Canadian_Stations.ipynb
+++ b/notebooks/PDSI_Canadian_Stations.ipynb
@@ -274,7 +274,7 @@
     }
    ],
    "source": [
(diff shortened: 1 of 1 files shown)


monocongo:
While this contribution is very appreciated and welcomed I should caution that the Palmers (PDSI, PHDI, etc.) are not scientifically vetted and not a reliable replication of the PDSI code used at NCEI.

monocongo:
I'm removing the Palmer indices from the upcoming 2.0.0 release. These were half-baked and nobody managed to take the ball and run with it, it just ended up adding a lot more effort for no real benefits.

PR #536, opened 2023-11-05
Update pyproject.toml

matteodefelice:
Recent xarray solved an issue that can affect the workflows based on climate indices.


monocongo:
It looks like this dependency update requires us to move up to Python 3.9 as the lowest supported version. Is there a Python 3.8 compatible version of xarray that fixes this issue instead? I'd like to keep Python 3.8 support if possible.

monocongo:
I have now updated the dependencies so this PR is now moot. Thanks for your help and attention on this one @matteodefelice, and apologies for such a late response!

PR #539, opened 2023-12-12
Palmer Drought Severity Index

nnayda:
@monocongo I had some time so I took the bait and worked on the Palmer Drought Severity Index. I took the existing [Fortran code](https://www1.ncdc.noaa.gov/pub/data/nidis/software/ncei/palmer/) from NOAA and parsed it to Python. I tried to keep the process flow the same and just change the syntax so that it would be easy to reference back. Therefore, the resulting code is not very pythonic. I think later on some enhancements and efficiencies could be gained, but at least it works.

I developed a suite of tests by running the Fortran code on all the available climate division data (downloaded from NOAA [here](https://www.ncei.noaa.gov/pub/data/cirs/climdiv/)), and then extracting the results. Some tweaks had to be made to the original Fortran code in order to get around differences in rounding errors. I stored these in fixture files to use in the pytests. These are the bulk of the file changes. I'm not sure what style preferences you have so I'm happy to restructure this - this just seemed the most logical layout to me.

Please let me know if I need to do anything else to support the pull request.


monocongo:
@nnayda This is huge, thanks so much for your effort. I'm very busy at my day job now but I should have time to review soon. THANK YOU!

monocongo:
I will add this to the next release on PyPI. Again, this is a significant contribution, the code here looks much better than the version I had before, and it seems to match the NOAA results. Thank you big time @nnayda 

BTW I'll eventually use this to process nclimgrid and various other large, gridded datasets to see how they look using this implementation. What are the other Palmer indices implementations and/or datasets in the wild that we can compare against next?

Issue #535, opened 2023-11-02
Xarray 2023.1.0 dependency


matteodefelice:
Good evening, I am using your fantastic library for a project but I need a recent version of xarray because of a bug that has been solved. I was wondering if you plan to move to a more recent version of xarray or include a less restrictive dependency.

monocongo:
Thanks for bringing this to my attention @matteodefelice

Yes, we can update the dependency to work around this bug. If you are comfortable doing so please update the `pyproject.toml` file accordingly and submit a PR. Otherwise please suggest a better version to use and I'll fix things myself. As long as the tests pass with a later version then we should be good to go.

matteodefelice:
Just done a PR, I hope that requiring xarray 2023.10.0 is not too strict, otherwise we can find something in between.

monocongo:
Seems like the PR will affect the ability to keep Python 3.8 support as that new version requires 3.9. What is the issue we're solving by using a more recent version of xarray? I'm happy to upgrade dependencies, but if we can do so without losing support for older Python versions that'll be preferable.

matteodefelice:
Unfortunately (or not), xarray dropped the support to Python 3.8 at the beginning of the year. The bug seems associated to pandas:
https://github.com/pydata/xarray/issues/2385

However, in my case, if I install xarray-2023.10 it disappears, reappearing when I go back to 2023.1 (pandas stays unchanged). 

I understand that you don't want drop Python 3.8, in that case I will try to find another version of xarray (or pandas) compatible with python 3.8 that doesn't trigger the error.

monocongo:
Thanks for sticking with it and looking for a 3.8-compatible solution.  It looks like numpy [dropped support for 3.8 this April](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) so we're also facing that. Eventually, we'll fall in line.

WeatherGod:
Why is the package pinning specific versions of dependencies? One should not be overly restrictive. Especially should not be specifying it down to the bugfix number, thereby preventing newer bugfix releases from getting used. These restrictions are preventing me from using newer versions of dask that supposedly fixed some chunking issues I'm having with the new ERA5 netcdf files.

monocongo:
Yes, I agree. Pinning to major versions should be sufficient.

WeatherGod:
And even that rule actually depends upon how that package does versioning. XArray's "major" version, for example, is the year of the release, but indicates nothing about compatibility.

[...]

PR #557, opened 2024-08-09
Relax package dependency requirements

WeatherGod:
* Relax dependency requirements to allow newer versions.
* Gain numpy 2.x compatibility by fixing np.NaN to np.nan

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,16 +44,16 @@ classifiers = [
 packages = [{include = 'climate_indices', from = 'src'}]
 
 [tool.poetry.dependencies]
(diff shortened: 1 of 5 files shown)


WeatherGod:
Closes #535. Trying it out without pinning numpy to <2.0 as, technically, it didn't before.

WeatherGod:
Just did a numpy 2.0 compatibility check with ruff, and there are some mistakes that looks like are easily fixable.

WeatherGod:
This also supersedes #536.

monocongo:
Thanks for the tidy-up for numpy, @WeatherGod

PR #560, opened 2024-09-19
Dependency version cleanups etc. for release of version 2.0.1

monocongo:
Edits for version updates, case-sensitivity of numpy NaN -> nan where used, and new synospsis docstring for main processing script

## Summary by Sourcery

Update the codebase for version 2.0.1 by enhancing type annotations and docstrings, refactoring the main processing logic, and updating test cases to use `np.nan` instead of `np.NaN`. Add a comprehensive synopsis docstring to the main processing script.

Enhancements:
- Add type annotations and docstrings to functions in the main processing script to improve code clarity and maintainability.
- Refactor the main processing logic into a new function `process_climate_indices` to separate concerns and improve readability.

Documentation:
- Add a detailed synopsis docstring to the main processing script, explaining the steps involved in processing climate indices.

Tests:
- Update test cases to use `np.nan` instead of `np.NaN` for consistency with numpy's preferred usage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ exclude = '''
 
 [tool.poetry]
 name = 'climate_indices'
-version = '2.0.0'
+version = '2.0.1'
 description = 'Reference implementations of various climate indices typically used for drought monitoring'
 authors = ['James Adams <monocongo@gmail.com>']
 readme = 'README.md'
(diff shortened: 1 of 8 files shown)


mvdebolskiy:
What's the reason for changing `scipy>=1.10` to `scipy==1.10.1`?

monocongo:
I have no recollection. What's your reason for asking? If there's a good reason it needs to be changed to another version number then please enlighten me. Thanks in advance!

mvdebolskiy:
@monocongo I am IT-support for a course, we have a setup with a quite complicated conda environment where `scipy>=1.13` is pinned, a student asked me to add this package, so I installed it with `-no-deps` and tested with the example repo (there is another problem with that (see #567). So I was just wondering why in this PR it was fixed to 1.10.1 instead of `>=1.10.1`.

monocongo:
That makes perfect sense. There are pros and cons to pinning specific version numbers, so maybe at the time I was opting to err on the side of caution? There have been instances in the past where the code around the L-moments and Pearson III distribution went wonky due to breaking changes in scipy, so that might have been what led me to that version pin. I'll do some testing with other versions, and feel free to do the same on your end if you have a list of versions that you think are better, as I'm happy to consider other options.

Issue #720, opened 2026-08-08
scPDSI PR4: implement palmer.scpdsi()


monocongo:
Part of epic #716. Depends on PR1, PR2, PR3.

Implement the public `palmer.scpdsi()` function per Wells, Goddard & Wilhite (2004), wiring together:
- Existing water-balance / CAFEC coefficient code (unchanged, reused from `pdsi()`).
- The generalized recursion from PR1, fed self-calibrated duration factors instead of the fixed ones.
- The statistics helpers from PR2 (duration-factor fitting, percentile rescaling).

**Algorithm** (see epic design doc for full derivation):
1. Compute unnormalized per-month K' factor (D/dbar-based, same formula as standard K but skip the final 17.67 normalization).
2. Compute raw Z-index for all periods (not just calibration years).
3. Fit wet/dry duration factors via PR2's rolling-window regression, restricted to the calibration period.
4. Run the PDSI recursion once with raw Z + fitted duration factors.
5. Iteratively rescale 3x: 2nd/98th percentile of calibration-period PDSI -> dry_ratio/wet_ratio -> rescale Z -> rerun recursion.
6. Derive scPHDI/scPMDI from the final recursion pass the same way `pdsi()` derives PHDI/PMDI.

**TDD**: write `tests/test_scpdsi.py` against PR3's fixtures first (red), then implement (green). Tolerance: start at `ATOL=5e-5` (matching `test_palmer.py`), widen only if a specific, explainable floating-point-path divergence is found.

**Acceptance criteria**:
- `palmer.scpdsi()` added to `__all__`, Google-style docstring, type hints.
- Passes against all 344 division fixtures within tolerance.
- `palmer.pdsi()` output is unaffected (regression check).

monocongo:
Two correctness considerations surfaced by PR1's (#717) final whole-branch review that become live once this issue starts feeding *fitted, per-location* duration factors through the generalized recursion (rather than PR1's identical wet==dry defaults, which mask them):

1. **`x3 == 0.0` tie-break disagreement.** `_select_duration_factors` treats `x3 >= 0` as wet, but `_calc_zindex`'s abatement dispatch routes `x3 == 0` to `_statement_180` (dry) for the `ze` computation. With PR1's defaults (wet factors == dry factors) this is invisible. Once scPDSI fits distinct wet/dry factors, this seam needs an explicit, documented choice — worth resolving as part of this issue's implementation rather than rediscovering mid-work.

[...]

Issue #719, opened 2026-08-08
scPDSI PR3: reference oracle + scPDSI fixture data


monocongo:
Part of epic #716.

Generate scPDSI/scPHDI/scPMDI/self-calibrated-Z-index reference values for the existing 344 climate divisions in `tests/fixture/palmer/`, for use as regression fixtures.

**Approach**: build a standalone C++ oracle by shimming the 4 Rcpp-specific calls (`NumericVector`, `NumericMatrix`, `Rf_error`, `Rf_warning`) in the vendored `Sibada/scPDSI` C++ source (`pdsi.cpp`/`pdsi.h`/`pdsi_ext.cpp`) with plain C++ equivalents. Compile with `clang++`. Run entirely outside the repo (session scratchpad) against the existing `precips.npy`/`pet.npy`/`palmer_awc.json` inputs (data start 1895, calibration 1931-1990, matching current fixture conventions).

**License constraint**: the reference source is GPLv3; this repo is BSD-3-Clause. The oracle build/run happens entirely outside the git repo. Only the resulting numeric fixture arrays are committed — no GPL source, headers, or derivative code enters the repository.

**Deliverables**:
- `tests/fixture/palmer/<div>/scpdsi.npy`, `scphdi.npy`, `scpmdi.npy`, `sczindex.npy` for all 344 divisions.
- `provenance.json` entry (or update) documenting the generation methodology, citing Wells, Goddard & Wilhite (2004) and noting cross-validation against the reference C++ implementation.

**Scope**: fixture data only. No `src/` changes.

monocongo:
Delivered in PR #744.

## What shipped

For all 344 divisions in `tests/fixture/palmer/<div>/`:
`scpdsi.npy`, `scphdi.npy`, `scpmdi.npy`, `sczindex.npy` (1536 float64 each), plus
`scdurfact.npy` — `[wetm, wetb, drym, dryb]`, one beyond the four series requested here. It
costs ~350 bytes per division and gives #720 a stage-level oracle for the duration-factor fit
rather than only an end-to-end target.

Provenance updated in `tests/fixture/palmer/provenance.json` (schema-validated), citing Wells,
Goddard & Hayes (2004) and documenting the methodology.

No `src/` changes, no test-code changes, 1067 tests passing.

## License boundary held

Built and run entirely in a session scratchpad. The branch contains **zero** `.cpp`/`.h`/`.sh`
[...]

Issue #717, opened 2026-08-08
scPDSI PR1: generalize PDSI recursion to accept duration factors


monocongo:
Part of epic #716.

Generalize `palmer.py`'s recursion functions (`_statement_170`, `_statement_180`, `_statement_190`, `_statement_200`, `_statement_210`) to read duration-factor slope/intercept from the `data` dict instead of hardcoding `0.897`/`3.0`, defaulting to today's Palmer-original values so `pdsi()`'s output is unchanged.

**Scope**: pure refactor. No new public API. No behavior change for `palmer.pdsi()`.

**Acceptance criteria**:
- All existing tests (1065) pass unchanged, including `tests/test_palmer.py`'s NOAA-fixture comparisons at the current tolerance.
- Duration factors are threaded through as explicit parameters/data-dict entries, not still hardcoded.
- `ruff check`, `ruff format --check`, `mypy` clean.

This is prerequisite for scPDSI PR4 (`palmer.scpdsi()`), which needs to reuse this recursion with self-calibrated duration factors instead of the fixed ones.

monocongo:
Done: statement_170/180/190/200/210 now read wetm/wetb/drym/dryb from the data dict (defaulting to Palmer's original 0.897/(1/3)-derived values). All existing tests pass unchanged at the existing ATOL=5e-5 tolerance; 7 new unit/integration tests in tests/test_palmer_duration_factors.py verify the parameterization is real and correctly wired. Ready for PR4 (#720) to feed self-calibrated duration factors through this same recursion.

Issue #716, opened 2026-08-08
Implement self-calibrating Palmer (scPDSI) per Wells et al. 2004


monocongo:
**Is your feature request related to a problem? Please describe.**

The library and CLI advertise a "Self-calibrated Palmer Drought Severity Index" (scPDSI) output (see `docs/algorithms.rst`, `docs/index.rst`, and the CLI's `--index palmers` output), but no self-calibration procedure actually exists in `src/climate_indices/palmer.py`. `palmer.pdsi()` only computes standard PDSI/PHDI/PMDI/Z-Index using fixed national duration factors and climate characteristic constants — there is no per-location recalibration of those constants.

Additionally, the CLI's Palmers dispatch path (`__main__.py`) was broken independently of this (missing `_parallel_process` branch, mismatched `func1d` return arity) — that plumbing bug was fixed separately (commit `1a4393b`) and the CLI's fake `scpdsi` output was dropped as part of that fix, rather than left in place as a misleading placeholder.

**Describe the solution you'd like**

Implement genuine self-calibration per Wells, Goddard & Wilhite (2004), "A Self-Calibrating Palmer Drought Severity Index":
- Per-location empirical calibration of the duration factors (used to determine when a wet/dry spell begins and ends)
- Per-location calibration of the climate characteristic (K), replacing the fixed constants currently used for the standard PDSI's Z-Index weighting
- A new public function `palmer.scpdsi()` returning the self-calibrated PDSI alongside standard PDSI/PHDI/PMDI/Z-Index
- Validation fixtures comparable to the existing `tests/fixture/palmer/` regression set, cross-validated against the Wells et al. reference C++ implementation

**Describe alternatives you've considered**

Leaving scPDSI undocumented/unadvertised until implemented — done as an interim step in commit `1a4393b` (removing the CLI's placeholder output, correcting docs). This issue tracks the actual follow-up implementation.

**Additional context**

Surfaced while building `src/climate_indices/CONTEXT.md` (domain glossary) and fixing the CLI's broken Palmers dispatch path. See the "scPDSI" entry in `src/climate_indices/CONTEXT.md` for the current-state note.

---

## Epic: this is tracked as 6 sub-PRs

There is no known definitive open-source Python implementation of scPDSI — this is real, substantial algorithm work (not a quick fix), broken into independently reviewable pieces:

- [x] #717 — PR1: generalize PDSI recursion to accept duration factors (pure refactor, behavior-preserving)
[...]

PR #742, opened 2026-08-08
scPDSI PR2: self-calibration statistics module

### LINKED ISSUE #718: scPDSI PR2: self-calibration statistics module
Part of epic #716.

New module (sibling to `palmer.py`) implementing the pure, independently-testable statistics pieces of Wells et al. (2004) self-calibration, ported from the reference C++ (`Sibada/scPDSI`, GPLv3 — used for algorithm reference only, no GPL source enters this repo):

- Rolling-window extreme Z-index sum per duration-factor window length (3,6,9,12,18,24,30,36,42,48 months), with the wet/dry asymmetric "reasonable extreme" filter from `get_Z_sum()`.
- Correlation-adaptive least-squares fit (`LeastSquares()`): trims trailing points until r >= 0.85, then adjusts the intercept through the most extreme residual point.
- kth-largest / percentile order statistic matching the reference's exact semantics (1-indexed, `k = int(percentage * size)`, NaN-safe).

**Scope**: pure functions, TDD'd against small hand-built/hand-computed arrays. Does NOT depend on the oracle/fixtures (PR3) or touch `palmer.py`.

**Acceptance criteria**:
- Each function has unit tests with known/hand-computed expected values.
- No dependency on `palmer.py` internals (imports one direction only: `palmer.py` will later import this module, not vice versa).

monocongo:
Part of epic #716, closes #718.

Adds `src/climate_indices/self_calibration.py` — the pure, independently-testable statistics pieces of the Wells, Goddard & Wilhite (2004) self-calibration procedure, ported from the reference C++ (algorithm reference only; no GPL source enters this repository):

- **`extreme_z_sum()`** — rolling-window extreme Z-index sum per duration-factor window length, including the wet/dry asymmetric "reasonable extreme" filter. The dry side returns the unfiltered extreme; the wet side discards sums whose ratio to the 98th-percentile rolling sum reaches 1.25. That asymmetry is a real feature of the reference, not a bug to normalize away.
- **`least_squares_fit()`** — correlation-adaptive least squares: trims trailing points until the sign-weighted correlation clears 0.85 (floor of 4 points), then anchors the intercept through the most extreme residual rather than through the sample means.
- **`kth_smallest()` / `nan_safe_percentile()`** — the reference's exact order statistic (1-indexed, truncating `k = int(fraction * size)`, NaN-safe), deliberately *not* `numpy.percentile`, whose interpolation would shift the self-calibration ratios.
- **`duration_factors()`** — composes the above over the monthly window-length set (3, 6, 9, 12, 18, 24, 30, 36, 42, 48 months) and normalizes by a PDSI value of ±4, yielding the `(m, b)` pair PR1 (#717) taught the recursion to consume as `wetm`/`wetb` or `drym`/`dryb`.

### Scope

Pure functions over numpy arrays. No `data`-dict machinery, no calibration-period bookkeeping (callers slice the Z series to the calibration window first), no xarray, no CLI, and no changes to `palmer.py`. A guard test enforces the one-way import edge — `palmer` → `self_calibration`, never the reverse — by walking the module's AST, which catches deferred and function-local imports as well as top-level ones.

### Testing

40 unit tests with hand-derived expected values, no fixture or oracle dependency. Full suite: 1101 passed, 37 deselected; `ruff` and `mypy` clean; `tests/test_palmer.py -m validation` unaffected.

The final review mutation-tested the module and found several mutations of the wet-side anomaly filter surviving the original test set, so the suite now explicitly pins `_REASONABLE_TOLERANCE`, `_EXTREME_PERCENTILE`, the strict `<` at the tolerance boundary, the inclusion of the initial-fill rolling sum, both missing-percentile sentinel branches, and the collinear-through-origin intercept anchor. Those are precisely the behaviours a well-meaning refactor would otherwise "fix" silently.

### Notes for review

Every function is a faithful port, including behaviour that looks wrong: the truncating order statistic, the wet-only outlier filter, and an intercept anchored on the most extreme residual (a perfectly collinear line through the origin yields `intercept = -slope * x[0]`, not 0). Each such quirk is documented inline with the reason it must be preserved — PR4 (#720) compares the whole pipeline against a C++ oracle at `ATOL=5e-5`, so accumulation order and rounding paths are load-bearing.

Degenerate-input guards are deliberately *not* added here: a constant `y` makes `least_squares_fit` raise (`ZeroDivisionError` when the constant is exactly zero, `ValueError: math domain error` otherwise, since sequential accumulation can leave `ss_y` slightly negative). That guard belongs with the `m + b == 0` guard already tracked on #720, where real fitted data flows through and the right behaviour can be decided against the oracle rather than invented.

### Dependencies

None — this PR touches no existing module. PR4 (#720) is the integration point that wires it together with #717 and #719.

## Summary by Sourcery

Introduce a new self_calibration statistics module for scPDSI with tests and guardrails, providing pure, numerically faithful building blocks for duration-factor fitting independent of palmer.py.

New Features:
- Add src/climate_indices/self_calibration.py implementing pure statistical functions for scPDSI self-calibration, including rolling-window extreme Z-index sums, adaptive least-squares fits, NaN-safe order statistics, and duration-factor computation.
- Expose duration-factor window length constants and wet/dry sign markers as the public API for self-calibration consumers.

Enhancements:
- Document the scPDSI self-calibration statistics design and implementation plan in a new superpowers plan file.
- Enforce the one-way dependency from palmer.py to self_calibration via an AST-based test that guards against import cycles.
- Define a clear __all__ surface for the new module and validate inputs using project-specific InvalidArgumentError rather than generic exceptions.

Tests:
- Add a dedicated test_self_calibration.py suite covering numerical behaviour, edge cases, and invariants of the new self_calibration functions, including anomaly filtering quirks and intercept anchoring semantics.

## Summary by CodeRabbit

* **New Features**
  * Added foundational self-calibration calculations for Palmer drought metrics, including percentile handling, extreme-value analysis, adaptive fitting, and duration-factor normalization.
  * Added robust handling for invalid inputs, missing values, and insufficient data.

* **Tests**
  * Added comprehensive coverage for calculations, edge cases, validation, numerical behavior, and public API boundaries.

* **Documentation**
  * Added an implementation plan outlining the self-calibration approach, testing strategy, validation requirements, and integration boundaries.


monocongo (commented):
# PR Review: #742 — scPDSI PR2: self-calibration statistics module

**Reviewed**: 2026-08-15
**Author**: monocongo
**Branch**: feature/scpdsi-2-selfcal-stats → main
**Decision**: APPROVE

## Summary

New, self-contained `src/climate_indices/self_calibration.py` module porting the Wells, Goddard & Wilhite (2004) self-calibration statistics (`kth_smallest`, `nan_safe_percentile`, `extreme_z_sum`, `least_squares_fit`, `duration_factors`) from the reference C++, with 41 hand-derived unit tests and an AST-based guard preventing `palmer.py` from being imported. Two follow-up commits (`0c4b0bb`, `c13d44e`) already on the PR head addressed the original review comments (deque/`popleft` instead of `list.pop(0)`, preallocated numpy array instead of a growing Python list for rolling sums, `math.isclose` with an exact-zero tolerance instead of `== 0.0`) and a SonarCloud finding. No further issues found.

## Findings

### CRITICAL
None

### HIGH
None

### MEDIUM
None

### LOW
None

## Validation Results

| Check | Result |
|---|---|
| Type check (`mypy`) | Pass |
| Lint (`ruff check`) | Pass |
| Tests (`pytest`, full suite) | Pass — 1102 passed, 37 deselected |
| Tests (`test_self_calibration.py` only) | Pass — 41/41 |
| Build | N/A (library package, no build step) |

## Files Reviewed

- `src/climate_indices/self_calibration.py` — Added
- `tests/test_self_calibration.py` — Added
- `docs/superpowers/plans/2026-08-08-scpdsi-pr2-self-calibration-stats.md` — Added (planning artifact)

## Notes

- Every quirky-looking behavior (truncating 1-indexed order statistic, wet-only anomaly filter, intercept anchored on the most extreme residual rather than the mean) is a faithful, well-documented port of the reference algorithm and is pinned by a dedicated test — verified against the inline rationale in each case, not just taken on faith.
- Degenerate-input guards (`m + b == 0`, constant-`y` regression inputs) are deliberately deferred to #720, where real fitted data will be available to decide correct behavior against the C++ oracle. This is called out explicitly in the code and PR description, not a silent gap.
- `InvalidArgumentError` usage matches the actual constructor signature in `exceptions.py` (positional `message`, keyword-only `argument_name`/`argument_value`/`valid_values`).
- The one-way `palmer` → `self_calibration` import boundary is enforced by an AST-walking test rather than a simple string grep, so it also catches function-local/deferred imports.

PR #744, opened 2026-08-09
test: add scPDSI reference fixtures and NCEI ground truth for 344 divisions

### LINKED ISSUE #719: scPDSI PR3: reference oracle + scPDSI fixture data

monocongo:
Closes #719. Part of epic #716.

Generates scPDSI / scPHDI / scPMDI / self-calibrated Z-index reference fixtures for all 344
climate divisions, plus — because validating them turned up a problem — the first genuine
external ground truth for the Palmer indices in this repository.

## What's here

| Path | Contents |
|---|---|
| `tests/fixture/palmer/<div>/sc{pdsi,phdi,pmdi}.npy`, `sczindex.npy` | 1536 float64 each, 344 divisions |
| `tests/fixture/palmer/<div>/scdurfact.npy` | 4 float64: `[wetm, wetb, drym, dryb]` |
| `tests/fixture/nclimdiv/{pdsi,phdi,pmdi,zindex}.npy` | NOAA NCEI operational values, `(344, 1536)` float32 |
| `tests/fixture/nclimdiv/divisions.json`, `provenance.json` | row index + provenance |
| `tests/fixture/palmer/provenance.json` | updated for both fixture groups |

`scdurfact.npy` is beyond the four series #719 asked for. It costs ~350 bytes per division and
gives PR4 a stage-level oracle: when `scpdsi()` diverges somewhere, PR4 can assert the
duration-factor fit in isolation rather than bisecting an end-to-end mismatch back through the
recursion and three rescaling passes.

No `src/` changes. No test-code changes. Full suite: 1067 passed.

## Licensing

The reference implementation (`Sibada/scPDSI`, Wells/Goddard/Hayes) is GPLv3; this project is
BSD-3-Clause. The oracle was built and run **entirely outside the repository** in a session
scratchpad, and only numeric arrays are committed — no GPL source, headers, shim code, or
build scripts. Verified: the branch contains zero `.cpp`/`.h`/`.sh` files.

The harness installed its four Rcpp shims (`NumericVector`, `NumericMatrix`, `Rf_error`,
`Rf_warning`) by include shadowing, so `pdsi.cpp` and `pdsi_ext.cpp` compiled **byte-identical
to upstream** (SHA-256 verified). The only edit to the reference was a 9-line additive
accessor in `pdsi.h` to feed inch-unit input without the millimetre round-trip — measured, that
round-trip perturbs 14% of P/PE values by 1 ULP, and scPDSI has order statistics and discrete
branch points where that can flip a result.

## Validation

**Harness** — running the same binary in standard-PDSI mode reproduces the committed
`zindex.npy` for **all 344 divisions at atol=5e-5**. Z-index is the product of the whole water
balance, the CAFEC coefficients and the K-factors, so this confirms units, AWC handling, the
calibration window and the text round-trip (which is separately proven bit-exact).

**The fixtures are genuinely self-calibrated** — independent of the reference implementation,
scPDSI's defining property holds: calibration-period 2nd/98th percentiles sit at
**−4.014 / +3.983 (sd 0.14 / 0.16)**, versus standard PDSI's **−4.55 / +4.38 (sd 0.83 / 0.91)**.
Self-calibration collapsed the cross-division spread ~6× onto ∓4, which is exactly what it is
supposed to do.

All 344 divisions produced finite, positive `m` and `b`, so the `m + b ≤ 0` blow-up flagged in
PR2 does not occur on this corpus.

## ⚠️ The finding reviewers should read

The validation gate **failed** on PDSI/PHDI/PMDI while passing on Z-index, and the reason is
architectural rather than a harness bug.

`palmer.py` is a port of **NCEI's Fortran `pdi.f`** — its functions are literally
`_statement_170` … `_statement_220`. `Sibada/scPDSI` is the **Wells, Goddard & Hayes** lineage.
They share the water balance but use different spell-selection state machines:
`palmer.py:330-333` resolves `x3 == 0` by largest absolute value, while the reference `ChooseX`
uses ±0.5 establishment thresholds with `Backtrack()` and explicit zero-cases. They disagree on
~6.3% of months by up to 8 PDSI units.

To settle which is right, this PR pulls the actual NOAA NCEI nClimDiv operational data. Over
528,384 division-months, restricted to the months where the two recursions actually disagree:

| series | disputed | climate_indices median err | Wells median err | climate_indices closer to NCEI |
|---|---|---|---|---|
| pdsi | 33,116 (6.3%) | **0.0143** | 0.1973 | **88.1%** |
| phdi | 39,057 (7.4%) | **0.0145** | 0.1719 | **87.9%** |
| pmdi | 279,616 (52.9%) | **0.0250** | 0.0334 | **65.1%** |

**Conclusions:**

1. `pdsi()`'s recursion is correct and must not change — swapping in the Wells recursion would
   make standard PDSI measurably worse.
2. But scPDSI *is* Wells et al. (2004), and NCEI publishes no self-calibrating PDSI. So
   `scpdsi()` must implement the Wells recursion. **The library needs two recursions,
   deliberately** — which contradicts PR1 (#717)'s premise of one generalized recursion serving
   both. Flagged on #720.
3. The epic design doc's "Verified consistency" section checked the `newX3`/`ZE` *formulas*
   (correct) but not the *control flow* around them (where they diverge).

## Also: the `noaa_*` fixtures are not NOAA data

`tests/test_palmer.py` loads the existing arrays into `noaa_pdsi` / `noaa_phdi` / `noaa_pmdi` /
`noaa_zindex`. Only **25.2%** of them fall within NCEI's own 0.005 rounding precision (median
deviation 0.0127) — copied values would agree to rounding. They are the library's own output,
exactly as `provenance.json` already said; the variable names are misleading.

Renaming them is deliberately left out of this PR to keep it free of test-code changes. The new
`tests/fixture/nclimdiv/` arrays are the real external reference, and are the first thing in
this repo that could validate `pdsi()` against something other than itself.

## Reviewing the numbers

The `.npy` files are opaque in a diff. Worth spot-checking:

```python
import numpy as np
d = "tests/fixture/palmer/0101"
sc, st = np.load(f"{d}/scpdsi.npy"), np.load(f"{d}/pdsi.npy")
print(np.median(np.abs(sc - st)))            # 0.1286 -- distinct, not a copy
print(np.load(f"{d}/scdurfact.npy"))         # [0.2588 3.2935 0.3358 3.2297]

# self-calibration property, calibration window 1931-1990
w = sc[(1931-1895)*12:(1991-1895)*12]
s = np.sort(w); n = w.size
print(s[int(0.02*n)-1], s[int(0.98*n)-1])    # ~ -4, ~ +4
```


monocongo:
## Adversarial self-review — two defects found and fixed

The planned Opus review agent died twice on infrastructure (a certificate error, then a
600s stall) without producing findings, so I worked the checklist inline instead. It found two
real problems in my own work. Both are fixed in `ac1ba4e` and `6acc58b`.

### 1. `sczindex.npy` was shifted by one month (all 344 divisions) — FIXED

```
zindex:   [ 1.4596, -2.6204,  0.8153, -1.8760, -0.9190, ...]
sczindex: [ 0.0000,  1.4324, -2.8083,  0.8001, -2.0105, ...]   <- before the fix
```

Each value held the *previous* month's Z, index 0 was a spurious `0.0`, and December 2022 was
lost entirely.

**Cause.** The reference is inconsistent about the period index it hands to `CalcOneX`.
`CalcOrigK` (standard path) does `month--` first, making it 0-based. `CalcZ`
(self-calibrating path) pushes `vals_mat(i,1)` — which `SumAll` wrote as
`per * period_length + 1`, so 1-based — straight into `PeriodList` with no adjustment.
`CalcOneX` indexes with `n = (year-1)*num_of_periods + period_number`, so every
self-calibrating write to columns 8–12 landed one row high, and the final period wrote at
`n == nrow()`, one past the end of the matrix (undefined behaviour in real Rcpp).

**Scope.** Only the Z-index is read from those columns, so only `sczindex.npy` was affected.
`scpdsi`/`scphdi`/`scpmdi` come from `Rext_output_X`, which uses its own sequential counter and
is correct on both paths — confirmed by the regeneration diff touching exactly 344 files, all
`sczindex.npy`.

**How it was caught.** Not by inspection — by an algebraic prediction. `sczindex/zindex` must
collapse to exactly **two** constants (the cumulative wet and dry rescaling ratios), because
the per-month K factor cancels:

```
Z_unnorm·cumratio / Z_norm = (d·k[m]·cumratio) / (d·17.67·k[m]/swtd) = cumratio·swtd/17.67
```

It was instead taking 1050 distinct values with 58.5% sign agreement. After the fix:

| metric | before | after |
|---|---|---|
| distinct ratios | 1050 | 2 clusters (0.98136, 1.07171) |
| within-branch relative spread, all 344 divisions | — | **8.5e-07** |
| sign agreement | 58.5% | **100%** |
| corr(sczindex, zindex) | 0.159 | **0.9997** |

Standard-mode output is unchanged and still reproduces the committed `zindex.npy` for all 344
divisions at atol=5e-5.

### 2. Committed docs reproduced four lines of GPL source — FIXED

The `Rext_use_inches` example in the spec and plan copied `Rext_init`'s soil-layer
initialization verbatim. Almost certainly de minimis, but the epic's constraint is that no GPL
source or shim code enters this BSD-3-Clause repo, and prose costs nothing. Replaced in
`6acc58b`.

## Checks that passed

- **`vals_mat` column mapping** (8=Z, 13=PDSI, 14=PHDI, 15=WPLM) verified against `pdsi.cpp`.
- **Duration-factor ordering** — a `drym`/`wetb` swap would silently corrupt all 344
  `scdurfact.npy`. The `--std` run discriminates: it emits `[0.309, 2.691, 0.309, 2.691]`,
  whereas a swap would give `[0.309, 0.309, 2.691, 2.691]`, since standard mode sets
  `wetm == drym` and `wetb == dryb`.
- **The `pdsi.h` edit** is purely additive (0 lines removed, 9 added). Auditing every use of
  `metric`, `AWC`, `Ss`, `Su`: `metric` is read on this path only in `Rext_get_Rvec`; the other
  five sites are in file-reading routines never entered. `SumAll()` re-derives the soil-layer
  split from `AWC` before any water-balance work, so correctness depends only on `AWC` holding
  the exact inch value.
- **Not an artifact** — 344/344 distinct `scpdsi.npy`, 344/344 distinct `scdurfact.npy`, no
  division duplicated, `corr(scpdsi, pdsi)` median 0.899 (range 0.692–0.991).
- **Duration factors are plausible** — converting to Palmer's form gives wet `p` ∈ [0.833,
  0.983] and dry `p` ∈ [0.845, 0.982], bracketing Palmer's fixed 0.897, with `q` bracketing 1/3.
  Exactly the regional variation self-calibration should produce.
- **`scphdi == scpdsi` on 72.5% of months**, against NCEI's own `pdsi == phdi` rate of 76.1%.
- **NCEI fixtures** — `divisions.json` matches the sorted fixture dirs; all 344 rows agree with
  the raw text for 1895; only element code `05` present in `pdsi.txt`; no `-99.99` survives; 0
  NaN; float32 round-trip error 4.6e-07.

## Statistics: no selection bias, but one honest caveat

Both headline numbers reproduce exactly (25.2%; 88.1%). The "disputed months" filter is
`|oracle − lib| > 5e-5`, which never references NCEI, so it cannot favour either candidate.

Worth stating plainly though: across **all** months it is **55.1% vs 44.9%**, not 88.1%. About
93.7% of months differ by less than 5e-5, where "closer to NCEI" is a coin flip driven by the
~0.013 noise floor from `precips.npy`/`pet.npy` not being byte-identical to NCEI's operational
inputs. The filter separates signal from noise rather than manufacturing a result — but the
88.1% figure should always be quoted as "on months where the two recursions materially
disagree", which is how it is worded in the provenance and on #716.

## Not completed

A full independent re-derivation of the Wells recursion in Python — implementing `CalcOneX`,
`ChooseX` and `Backtrack` from scratch and reproducing `scpdsi.npy` end to end. `Backtrack`
rewrites already-emitted values and is substantial. That work is #720's job, and #720 will
reproduce these fixtures or fail loudly, which is the check that matters.

monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/744 (`feature/scpdsi-3-oracle-fixtures` → `main`, head `ef1cb5c`)
- **Lanes run:** 4 — data correctness/integrity, licensing/compliance, test-suite validation, scope/maintainability
- **Validation:** `uv run pytest -q` → 1067 passed, 37 deselected (matches PR's claimed count); independent Python re-derivation of every quantitative claim in the PR description and both `provenance.json` files (checksums, self-calibration collapse statistics, NCEI agreement percentages) — all reproduced exactly
- **Fix commits:** None — no verified defect required a code change

### Findings and fix status

No P0–P3 findings. This PR is fixture data only (1731 files: 1727 `.npy`, 3 `.json`, plus `AGENTS.md`, two `docs/superpowers` files, and `uv.lock`) — there is no `src/` or test-code diff to review for logic defects, so the review focused on verifying the data itself and the PR's own claims about it, all of which check out:

- All 344 `tests/fixture/palmer/<div>/sc{pdsi,phdi,pmdi}.npy` and `sczindex.npy` are `(1536,)` float64, finite, no NaNs.
- All 344 `scdurfact.npy` are `(4,)`, all four values positive, `m + b > 0` on both wet and dry branches for every division (confirms the PR's claim that the PR2-flagged blow-up condition doesn't occur on this corpus).
- `tests/fixture/nclimdiv/{pdsi,phdi,pmdi,zindex}.npy` are `(344, 1536)` float32 as documented; `divisions.json` row order matches the `tests/fixture/palmer` directory listing exactly.
- Both provenance checksums are exactly reproducible: `nclimdiv/provenance.json`'s `checksum_sha256` (concatenated sorted per-file SHA-256 digests) and `palmer/provenance.json`'s (division 0101, sorted `.npy` concatenation) both match a from-scratch recomputation.
- Self-calibration collapse statistic reproduces to 3 decimals: scPDSI calibration-window 2nd/98th percentiles −4.014/+3.983 (sd 0.141/0.161) vs standard PDSI −4.547/+4.384 (sd 0.827/0.911) — exact match to the PR body and provenance notes.
- The `noaa_*`-fixture-vs-NCEI agreement numbers in `nclimdiv/provenance.json` are independently reproducible: pdsi median 0.0127 abs diff, 25.2% within NCEI's 0.005 rounding precision, 86.2% within 0.05; phdi 20.6% within 0.005; pmdi 17.9%; zindex 24.9% — all match.
- No stray file types leaked in: every changed path is `.npy`/`.json`/`.md`, plus `AGENTS.md` and `uv.lock`. No `.cpp`/`.h`/`.sh`/build scripts, confirming the PR's GPLv3-boundary claim. No `src/` changes, no non-fixture `tests/` changes, confirming "no src changes, no test-code changes."
- The two new `docs/superpowers/*` files contain no verbatim reference source — a prior commit on this branch (`6acc58b`, already part of this PR head) had caught and removed 4 lines of GPL-derived C++ that had been reproduced in an earlier draft of the plan/spec, replacing them with a prose description. Current head is clean of literal reference source; only type/function *names* (e.g. `Rcpp::NumericVector`) appear, describing the shim's own API surface rather than reproducing copyrightable expression.
- Full test suite passes unchanged (1067 passed, 37 deselected) — expected, since no fixture is wired into test code yet (that's deferred to a future PR per the description).

**Non-blocking observation (not filed as a numbered finding — no file/line defect, just a scope note):** `AGENTS.md` (new "Git safety" section) and `uv.lock` (routine `jupyterlab`/`pillow` patch bumps) are unrelated to this PR's stated fixture-data purpose. Both are individually correct and harmless, so there's nothing to fix, but bundling them in makes the PR slightly harder to review/revert atomically. Consider a follow-up PR carrying only those two changes if that matters to your workflow — optional.

### Remaining blockers
None.

Issue #740, opened 2026-08-08
Harden flaky PET Hargreaves xarray overhead benchmark


monocongo:
> *This was generated by AI during triage.*

## Summary

The PET Hargreaves xarray overhead gate is flaky on GitHub-hosted runners. Normal runner variation moves the measured overhead across the hard 80% threshold even when the tested code and dependency lock are unchanged.

## Evidence

- PR #736 failed at **84.9%** overhead (NumPy 1.8 ms, xarray 3.4 ms), then passed when the failed job was rerun without code changes: https://github.com/monocongo/climate_indices/actions/runs/31271177557
- An unrelated documentation PR failed the same test at **84.7%** on the same day: https://github.com/monocongo/climate_indices/actions/runs/31260936694
- A July workflow failed the same test at **85.7%**: https://github.com/monocongo/climate_indices/actions/runs/29922399016
- Six nearby successful CI artifacts measured pytest-benchmark mean ratios between **70.9% and 78.2%**, leaving only about two percentage points of typical headroom.
- Repeated isolated and full-suite runs under Python 3.12 passed locally.
- PR #736 had no changes to PET, the xarray adapter, benchmark code or fixtures, dependencies, or the benchmark workflow.

## Current behavior

The overhead check independently measures the NumPy and xarray callables with `timeit.repeat`, selects each minimum, computes `(xarray - numpy) / numpy`, and compares every operation to one 80% limit. Hargreaves normally runs close enough to that limit on CI that routine host variation can change the result from pass to fail.

The rerun's pytest-benchmark means were 2.65 ms for NumPy and 4.56 ms for xarray (about 72% overhead), confirming that the original failure did not represent a persistent regression.

## Desired behavior

Keep a meaningful CI performance gate, but make its measurement and budget robust enough to distinguish a real xarray adapter regression from ordinary GitHub-hosted-runner variation. Document the rationale and retain useful timing output when the gate fails.

Do not address this by changing Hargreaves PET calculations or by broadly weakening unrelated performance checks without supporting measurements.

monocongo:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug  
**Summary:** Make the PET Hargreaves xarray overhead CI gate resistant to normal hosted-runner variance while preserving regression detection.

**Current behavior:**
[...]

PR #757, opened 2026-08-28
fix(xarray): make calendar semantics explicit at the xarray seam

monocongo:
## Summary

Makes calendar semantics explicit at the xarray seam — the top recommendation from the architecture deepening review merged in #753 ([`docs/architecture-deepening-review-2026-08-27.md`](docs/architecture-deepening-review-2026-08-27.md), candidate 1).

The NumPy index cores group monthly values into 12 positional slots per year and daily values into 366, treating every year as a leap year beginning in January. The xarray adapter passed raw Gregorian values straight through and then reattached the original coordinates — so daily results were shifted for every value after February 28 of a non-leap year, and reattaching the coordinates hid the drift. The CLI already owned this conversion via `utils.transform_to_366day()` / `utils.transform_to_gregorian()`; the xarray seam did not.

Conversion and validation now live in one place, `_build_daily_calendar_plan()`, applied both by the `@xarray_adapter` decorator and by the hand-written PET paths. The NumPy implementations are untouched, per [ADR-0001](docs/adr/0001-dual-numpy-xarray-api.md), and the separate execution paths of [ADR-0002](docs/adr/0002-multiprocessing-cli-dask-xarray.md) are preserved.

## Behavior changes

**Daily xarray results change numerically** for any series spanning a non-leap year — they were drifted and are now calendar-aligned. Affects `spi()`, `spei()`, `eddi()`, `percentage_of_normal()`, and `pet_hargreaves()`. **NumPy API results are unchanged.**

**Stricter time-coordinate validation.** Inputs previously accepted and silently misinterpreted now raise `CoordinateValidationError`:

- `cftime` calendars — only `standard`, `gregorian`, and `proleptic_gregorian` `datetime64` coordinates are supported
- monthly series not beginning in January; daily series not beginning on January 1
- a coordinate whose inferred periodicity does not match the requested one (e.g. daily data passed to `pet_thornthwaite()`)

A partial *final* year remains supported. A partial *first* year is rejected rather than normalized — padding one would require guessing where the caller's calendar begins, and guessing wrong reintroduces the silent drift this change removes.

## Design decisions

Recorded in [ADR-0004](docs/adr/0004-xarray-calendar-semantics.md), which resolves the open questions from the implementation handoff: Gregorian-only for the first slice, reject rather than normalize non-January origins, keep partial final years, and own conversion at the seam rather than pushing calendar knowledge into the NumPy cores.

## Test plan

- [x] `uv run ruff check src/ tests/` — clean
- [x] `uv run ruff format --check src/ tests/` — clean
- [x] `uv run mypy src/` — clean
- [x] `uv run pytest` — 1154 passed, 37 deselected
- [x] `uv run pytest -m benchmark` — 20 passed, 4 skipped
- [x] `notebooks/xarray_getting_started.ipynb` and `notebooks/eddi_xarray.ipynb` execute cleanly
- [x] End-to-end: daily Hargreaves over 2019 (non-leap) → 2020 (leap) matches explicit `transform_to_366day` → `eto_hargreaves` → `transform_to_gregorian`, diverges from the raw positional result at exactly index 59 (March 1, 2019), and preserves the input time coordinate

New regressions in `tests/test_xarray_equivalence.py` compare the public xarray interface against explicit CLI-style conversion for daily SPI, PNP, EDDI, SPEI, and PET Hargreaves, covering a non-leap → leap sequence, a partial final year, Dask-backed input under the one-time-chunk invariant, and each rejection path.

`test_pet_hargreaves_equivalence` previously asserted that the xarray path equalled the raw positional NumPy path — it pinned the bug. It now compares against CLI-style conversion, with a companion test asserting the result diverges from the raw positional output at March 1.

## Notes for the release

No `CHANGELOG.md` entry: `test_release_integrity.py::test_changelog_has_no_unreleased_block` forbids an `## [Unreleased]` block, and `test_changelog_top_entry_matches_pyproject_version` pins the top block to pyproject's version. The two behavior-change sections above are written to be folded into the next release block.

`_PET_THORNTHWAITE_OVERHEAD_THRESHOLD` is a new operation-specific benchmark budget, following the existing `_PET_HARGREAVES_OVERHEAD_THRESHOLD` precedent (#740). The added calendar check costs one `xr.infer_freq` call (~0.14 ms) against Thornthwaite's ~0.4 ms NumPy baseline — the cheapest operation measured — moving the ratio from under 80% to 85–98%. The cost is fixed per call and amortizes to nothing on gridded data.

`pci()` is not covered by the calendar contract: it is a manual single-year wrapper requiring exactly 365 or 366 values, and its semantics warrant a separate decision. Neither is `@xarray_adapter(infer_params=False)`, which disables coordinate inference and with it the calendar contract — no shipped wrapper uses it, but it is the one remaining path that can still compute positionally. Both are recorded in the ADR consequences.

## Review round

A code review against the project rules and ADRs produced one blocking finding and several advisories, addressed in `fix(xarray): address code review findings on the calendar seam`:

- `_DailyCalendarPlan`'s shape guards raised bare `ValueError`, which `.claude/rules/architecture.md` prohibits. They now raise `DataShapeError` with `expected_shape` / `actual_shape`, with tests asserting the subclass and its context attributes, plus a round-trip case over a partial final year.
- `_validate_supported_calendar` read the calendar name only from `.attrs`; xarray records it in `.encoding` when opening a dataset. It now consults both.
- Added negative controls asserting the CLI-style references for SPI, SPEI, EDDI, and PNP genuinely differ from the raw positional results, so the calendar comparisons cannot pass vacuously on data that never straddles a February 29 boundary.
- ADR-0004 claimed full coverage; it now records the `infer_params=False` carve-out.

The reviewer independently verified that `to_gregorian(to_all_leap(x)) == x` exactly across leap and non-leap start years at lengths spanning the Feb 28 / Feb 29 / Mar 1 boundaries, and that `to_all_leap` matches `utils.transform_to_366day` on whole-year input.

### Follow-ups not taken here

Filed as #758, #759, and #760.

- **#758** — `_PET_THORNTHWAITE_OVERHEAD_THRESHOLD` is a ratio budget, which at 120% permits roughly 2.2x the NumPy path. Since the added cost is a fixed per-call `xr.infer_freq`, an absolute budget would guard the real invariant more precisely — worth applying consistently across all operations rather than to this one in isolation.
- **#759** — `_resolve_periodicity` fails open (returns `None`, meaning no validation) when periodicity is not a `Periodicity` instance. Not currently reachable — `indices.spi` rejects a string periodicity with `InvalidArgumentError` before any result escapes — but a fail-open default inside the mechanism that prevents silent failure is worth revisiting.
- **#760** — Calibration sample-size validation runs on the pre-conversion Gregorian values while the fit runs on the 366-day array, so `ShortCalibrationWarning` thresholds are computed against slightly different counts than the fitter uses.

## Summary by Sourcery

Make calendar semantics explicit at the xarray boundary so daily results honor Gregorian dates while unsupported or ambiguous coordinates fail clearly.

Bug Fixes:
- Align daily xarray calculations with Gregorian calendar semantics by converting through the NumPy core’s 366-day layout and restoring the original dates.
- Reject unsupported calendars, non-January origins, and periodicity mismatches instead of silently producing positionally misinterpreted results.

Enhancements:
- Centralize calendar validation and conversion across decorated xarray operations and handwritten PET adapters while preserving the NumPy APIs and Dask execution model.
- Improve periodicity detection performance for supported daily and monthly layouts and extend validation across secondary inputs.

Documentation:
- Document the xarray calendar contract and record its design decisions in ADR-0004.

Tests:
- Add equivalence and regression coverage for calendar-aligned daily SPI, SPEI, EDDI, PNP, and Hargreaves PET results, including Dask inputs, partial final years, rejection paths, and shape errors.

Chores:
- Update benchmark fixtures and expectations for calendar-aware daily processing.

## Summary by CodeRabbit

* **New Features**
  * Daily xarray calculations now preserve Gregorian calendar alignment, including synthesized February 29 values and restored output coordinates.
  * Partial final years are supported, with padded values removed from results.
  * Added validation for supported calendars, January-based input, periodicity, shapes, and Dask inputs.

* **Bug Fixes**
  * Corrected daily result shifts after February 28 in non-leap years.
  * Updated PET and climate-index calculations to produce calendar-aware results.
  * Unsupported calendars and invalid date origins now raise clear validation errors.

* **Documentation**
  * Added calendar compatibility guidance, migration instructions, and an architectural decision record.
  * Documented unsupported calendar types and updated affected API guidance.


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/757 (`fix/xarray-calendar-semantics` → `main`)
- **Lanes run:** 3 independent agent lanes (calendar-conversion numerical correctness; Dask and multi-array/secondary-input path correctness; test quality and ADR/docs consistency) plus direct orchestrator verification
- **Validation:** `uv run ruff check src/ tests/` — pass; `uv run ruff format --check src/ tests/` — pass; `uv run mypy src/` — pass; `uv run pytest -m "not benchmark and not validation"` — 1207 passed, 0 failed; `uv run pytest -m benchmark` — 20 passed, 4 skipped; `tests/test_review_scripts.py` (llms bundle sync) — 13 passed; `sphinx-build -b html docs` — no new warnings
- **Fix commits:** [`b9a667d` — docs: correct beta-feature no-breaking-changes claim for calendar fix](https://github.com/monocongo/climate_indices/commit/b9a667d), [`acb1ffb` — docs: distinguish all_leap rejection rationale in calendar ADR](https://github.com/monocongo/climate_indices/commit/acb1ffb)

This PR's core contribution — correcting the xarray adapter's silent daily-calendar drift for non-leap years — held up under independent, executed verification. All three lanes and the orchestrator ran real code (not just read it) to check the load-bearing claims, and found no P0/P1 correctness defects.

### Findings and fix status

#### [P2] Beta-feature documentation directly contradicted this PR's own stated behavior change — Fixed
- **Location:** `docs/xarray_migration.rst:13-19` (the "Beta Feature" warning box)
- **Origin:** Orchestrator (sharpened from a broader observation in the test-quality/ADR-consistency lane, which flagged that the migration docs described the new *exceptions* well but didn't surface the no-error numerical-output change)
- **Problem:** The existing warning box stated: "The underlying computation results are identical to the stable NumPy API. No breaking changes will occur within a minor version." This PR's own description says the opposite: daily xarray results for `spi()`, `spei()`, `eddi()`, `percentage_of_normal()`, and `pet_hargreaves()` change numerically for any series spanning a non-leap year, with no error raised — and it ships as a minor version bump (`pyproject.toml` is at `2.4.0`; `CHANGELOG.md` references `v2.5.0` as the next release). A user relying on the documented guarantee would have no reason to expect their pipeline's output to change silently on upgrade.
- **Applied fix:** Softened the original claim to reflect intent rather than an unconditional guarantee, and added an adjacent warning box explicitly naming the affected functions, the no-error behavior, and a pointer to ADR-0004 and the new "Unsupported calendar or calendar origin" pitfall entry; that same pitfall entry's "Cause" text now also names `pet_thornthwaite`/`pet_hargreaves` as enforcing the same contract, alongside SPI/SPEI/EDDI/PNP. Regenerated `llms-full.txt` (which bundles `docs/xarray_migration.rst`) to keep `tests/test_review_scripts.py::test_llms_bundle_matches_configured_sources` passing. Commit `b9a667d`. A follow-on commit, `acb1ffb`, separately clarifies the ADR's rationale for rejecting `all_leap` cftime calendars (they match the 366-day layout exactly but are rejected because the interface only accepts `datetime64`, not cftime).
- No CHANGELOG.md change was made: the repo enforces (`tests/test_release_integrity.py::test_changelog_has_no_unreleased_block`) that no `## [Unreleased]` block exists outside of a release cut, and the PR's own "Notes for the release" section already explains this is deliberately deferred to release time — not a gap in this PR.

Investigated and found sound, no fix needed (each independently verified by execution, not just inspection):
- **`_DailyCalendarPlan.to_all_leap`/`to_gregorian` round-trip correctness** — hand-computed and executed against 13 cases (full/partial years, leap/non-leap, multi-year chains); all exact.
- **Non-contiguous-but-correct-length daily data** — constructed two adversarial cases (a compensated gap, a duplicate-plus-dropped date); both correctly rejected by the existing `xr.infer_freq`-based periodicity check before any plan is built.
- **Feb 29 synthesis arithmetic** (`(source_year[58] + source_year[59]) / 2`) — confirmed index 58/59 are Feb 28/Mar 1 via `pd.date_range`.
- **Dask detection and per-array chunk validation across primary + secondary inputs** — confirmed `is_dask` correctly flips when only a secondary is Dask-backed, and mismatched non-core chunk layouts between primary and secondary do not corrupt results (Dask auto-unifies broadcast-dim chunks).
- **Calendar-plan threading through `apply_ufunc` wrappers and `pet_hargreaves`'s hand-written wrapper** — the uniform "every positional array is a time series" assumption in `_make_calendar_aware_numpy_wrapper` is safe because `_validate_calendar_secondary_inputs` rejects any non-`xr.DataArray` secondary before that point, and the only shipped secondary with a time dimension (SPEI's `pet_mm`) is genuinely a time series; `_hargreaves_with_copy`'s `{0,1,2}`/lat-at-3 split matches `eto.eto_hargreaves`'s real signature.
- **ADR-0004 vs. implementation** — every concrete claim (supported calendars, January-origin rules, partial-final-year support, the `infer_params=False` carve-out, the `pci()` exclusion) checks out against the code.
- **Exception usage** — all new `CoordinateValidationError`/`InputTypeError`/`DataShapeError` call sites match their constructors in `exceptions.py`; no bare `ValueError`/`RuntimeError` introduced.
- **New test coverage** — proper `assert_allclose`/`assert_equal` usage, non-vacuous negative controls, and genuine Dask-backed-secondary test coverage (not just Dask-backed primary).
- **New `_PET_THORNTHWAITE_OVERHEAD_THRESHOLD` benchmark budget** — reproduced locally (~106% against a 120% budget), consistent with the existing Hargreaves precedent.

Noted but out of scope for this diff (pre-existing, not introduced by this PR, and not currently reachable via any shipped index):
- A non-time-series secondary `DataArray` without a `time` dim surfaces a raw xarray `ValueError` from `apply_ufunc`'s internals rather than a `ClimateIndicesError`, in the multi-dimensional path only.
- No runtime check rejects a `latitude` argument that itself carries a `time` dimension in `pet_hargreaves`.

### Remaining blockers
None. All findings are fixed and verified at final HEAD; lint, format, mypy, the default suite, the benchmark suite, the llms-bundle sync check, and a full Sphinx doc build all pass.

monocongo:
@sourcery-ai review

PR #821, opened 2026-09-13
chore: remove commented-out pearson_parameters_previous

marekl11:
`compute.py` carries a fully commented-out copy of an older Pearson Type III fitting routine, `pearson_parameters_previous`, sitting between the live `pearson_parameters()` and `_minimum_possible`. All 121 lines of it are comments, so nothing imports or runs it, and git history keeps the original if it is ever wanted back.

This removes the block and nothing else. The `# +++` section divider above it and both neighbouring functions are untouched.

Checked before opening:

- no reference to the name anywhere in the tree (source, tests, docs, generated llms bundles)
- `ruff check src/ tests/` and `ruff format --check src/ tests/` clean
- `mypy src/` clean
- `pytest -m "not benchmark and not validation"`: 1245 passed, 3 skipped

## Summary by Sourcery

Enhancements:
- Remove the obsolete commented-out Pearson Type III fitting routine from the computation module.

## Summary by CodeRabbit

* **Chores**
  * Removed obsolete commented-out code from the codebase.
  * No user-facing functionality or behavior changed.


monocongo:
Thanks @marekl11 good catch/cleanup

PR #845, opened 2026-09-14
test: read repository files as UTF-8 on Windows

mclainv:
## Summary

Pass `encoding=utf-8` to every bare `Path.read_text()` in `tests/test_release_integrity.py` and `tests/test_pattern_compliance.py`.

On Windows, `Path.read_text()` uses the locale encoding (`cp1252` here). `README.md` contains a curly quote whose UTF-8 third byte is `0x9D`, which `cp1252` cannot decode, so `uv run pytest` fails before the rest of the suite matters. CI only runs on Ubuntu and macOS, so this never shows up there.

Two calls in `test_release_integrity.py` already passed `encoding=utf-8`; the rest of that file and `test_pattern_compliance.py` now match. `CHANGELOG.md` also has non-ASCII bytes that `cp1252` would silently mojibake.

## Test plan

- [x] `uv run pytest tests/test_release_integrity.py tests/test_pattern_compliance.py` — 111 passed on Windows
- [x] `uv run ruff check src/ tests/`
- [x] `uv run ruff format --check src/ tests/`
- [x] `uv run mypy src/`
- [x] `uv run pytest` — 1223 passed on Windows (was 1222 passed, 1 failed)

Made with [Cursor](https://cursor.com)

## Summary by Sourcery

Bug Fixes:
- Ensure repository files are read as UTF-8 in integrity and pattern-compliance tests, preventing Windows locale encodings from causing test failures or corrupted text comparisons.

## Summary by CodeRabbit

* **Tests**
  * Standardized test file reading to use UTF-8 encoding, improving consistency across environments.
  * Consolidated release workflow validation into shared file-reading logic while preserving existing coverage.
  * Continued validating release configuration, Python version support, dependency handling, packaging, and publishing requirements.


monocongo:
Thank you @mclainv !

PR #966, opened 2026-09-16
refactor(palmer): share one duration-factor contract and one CAFEC/Z-index step (#899)

monocongo:
## Summary

Steps 1-2 of the phased solution in #899: the CAFEC/Z-index arithmetic and the
duration-factor `(m, b) -> coefficients` contract each get one owner.

- `palmer._calc_cafec_zindex()` owns the one-month CAFEC and raw Z-index arithmetic
  that the standard PDSI (`_calc_zindex`) and scPDSI (`_calc_scpdsi_raw_zindex`)
  recursions had copy-pasted.
- New private module `src/climate_indices/_palmer_duration.py` owns the
  duration-factor contract:
  - `DurationFactors.from_defaults()` — Palmer's (1965) fixed national factors,
    derived from the published `p`/`q` constants (moved out of `palmer.py`).
  - `DurationFactors.from_fitted()` — validation plus the Wells recurrence
    coefficients, moved verbatim from `_palmer_wells._validated_factors`;
    `_palmer_wells._Factors` is now this type.
  - `DurationFactors.weighting_fraction(m, b)` — the pdi.f lineage's
    `c = b / (m + b)`, moved verbatim from `palmer._duration_factor_c`.

## The two coefficient expressions are deliberately not merged

`1 - m / (m + b)` (Wells) and `b / (m + b)` (pdi.f) agree in exact arithmetic but
differ in the last bit of the float, and both recursions branch on exact comparisons
(`_is_exact_zero`, `-0.5 <= x3 <= 0.5`, `z >= 0.15`). So each lineage keeps its own
expression inside the one owner, with the reason recorded in the module docstring.

Evidence that the refactor is arithmetic-neutral: comparing the refactored paths
against verbatim pre-refactor arithmetic over all 344 division fixtures, `cp` and `z`
are bit-identical month by month.

## Scope decisions

- Step 3 of #899 (typed prepared-input/output structs replacing the ~50-key `data`
  dict) is **not** in this PR. It touches every `_calc_*`/`_statement_*` helper plus
  the tests that mutate dict keys mid-pipeline, and the issue grades it "Worth
  exploring" against "Strong" for steps 1-2. #899 stays open for it.
- Merging the two recursion engines stays out of scope per the issue; it waits on
  #755 and per-path oracle gates.
- No behavior change intended, and none observed: the `data` dict keys, their values,
  and every public signature are unchanged.

## Verification

- `uv run ruff check src/ tests/` — clean
- `uv run ruff format --check src/ tests/` — clean
[...]

Issue #899, opened 2026-09-16
TIDY_UP-06: Palmer shared-stage deepening: CAFEC/z helper, one duration-factor contract


monocongo:
## Files
- `src/climate_indices/palmer.py` — `_prepare_palmer_data` (:979-1064), `_calc_zindex` (:724-795), `_calc_scpdsi_raw_zindex` (:363-381), `_duration_factor_c` (:44-58), `_initialize_data` (:863-958)
- `src/climate_indices/_palmer_wells.py` — duration-factor coefficients (:120-135), recursion engine (:157-368)
- `src/climate_indices/self_calibration.py` — `duration_factors` (:404-453)

## Problem
`_prepare_palmer_data` is the one real shared seam between standard PDSI and scPDSI; downstream of it the same arithmetic is written twice, and the module's de facto interface is an untyped ~50-key `data` dict mutated by `_calc_*`/`_statement_*` helpers. Concrete duplications:
- CAFEC/z arithmetic is identical at `palmer.py:367-373` and `palmer.py:743-749` (still present after #856);
- duration-factor coefficients are computed as `b/(m+b)` in `palmer.py:44-58` and as `1 - m/(m+b)` in `_palmer_wells.py:120-135`;
- two recursion engines (`palmer.py:406-820` vs `_palmer_wells.py:157-368`) encode the same Wells semantics; `tests/fixture/palmer/provenance.json` records the lineages disagreeing on ~6.3% of months, so merging them is not a safe first step.

## Solution (phased)
1. Extract one CAFEC/z helper over the prepared arrays, called by both `_calc_zindex` and `_calc_scpdsi_raw_zindex` (numeric-identical).
2. One `DurationFactors` value object (or coefficient function) owning the `(m, b) → coefficients` contract, constructed from defaults or `self_calibration.duration_factors`.
3. Typed prepared-input/output structs replacing the mutated dict, stage by stage.

Out of scope: merging the two recursion engines — that needs per-path oracle gates and should wait for the scPDSI follow-ups (#755).

## Benefits
Locality: one arithmetic path for both Palmer variants. Testability: stages become callable without replaying private `_calc_*` chains (`tests/test_palmer_duration_factors.py:175-209`).

## Recommendation strength
**Strong** for steps 1-2; **Worth exploring** for step 3. Recursion-engine merge is **Speculative** until #755 closes.

## Related
[...]

PR #969, opened 2026-09-16
test(palmer): harden duration-factor and CAFEC bit-identity pins

monocongo:
## Summary

Follow-up hardening for #966 (steps 1-2 of #899), which merged while review was still in flight. The refactor itself proved behavior-neutral; the findings were tests that claimed to pin the refactored contracts but could not fail if those contracts were silently changed.

- `test_fitted_coefficients_keep_the_wells_complement_form` used constants where the Wells complement form and the pdi.f division form are bit-identical for `wetc` and `dryc`; now uses `0.1, 0.2, 0.1, 0.3`, where all three coefficients differ bitwise between forms and the `drym + wetb` cross denominator is still pinned.
- `test_weighting_fraction_keeps_the_division_form` had the same blind spot at `1.7, 2.3`; now uses `0.1, 0.2` plus an explicit `!=` against the complement form.
- `test_calc_cafec_zindex_writes_cafec_and_zindex` asserted with `pytest.approx` on exactly representable values, so a 1-ulp reassociation of the CAFEC sum, or distributing `ak` over the departure, was invisible even though the recursions branch on exact comparisons. It now compares exact equality against the pre-refactor named-intermediate grouping, with constants that change the last bit under those reassociations.
- `test_fitted_duration_factor_coefficients_stay_contractions` recomputed the three coefficients with test-local formulas instead of calling `DurationFactors.from_fitted`; it now exercises the production derivation over all 344 divisions.
- The invalid-denominator parameterization matched `"duration factors"`, which the coefficient-magnitude message also satisfies; the match is now anchored to `recursion$`, so the exact-zero cross-denominator branch is pinned as such.
- Documentation overstated that the two coefficient forms always differ bitwise; changed to "may differ" / "neither is guaranteed bit-identical".

The `src/climate_indices/_palmer_duration.py` and `CONTEXT.md` changes are wording only; no behavior or public API changes.

## Verification

- `uv run ruff check src/ tests/` - clean
- `uv run ruff format --check src/ tests/` - clean
- `uv run mypy src/` - clean
- `uv run pytest` - 1972 passed, 406 deselected
- `uv run pytest -m validation` - 378 passed, 3 skipped
- `uv run scripts/generate_llms_txt.py` - no bundle drift

## Related

- #966 - merged refactor this follows up on
- #899 - Palmer shared stages (steps 1-2 merged)

## Summary by Sourcery

Harden Palmer numerical contract tests and clarify floating-point behavior without changing runtime behavior or the public API.

Enhancements:
- Harden Palmer duration-factor and CAFEC/Z-index contract tests so changes to floating-point formulas, exact-value grouping, and fitted-coefficient derivation are detected reliably.

Documentation:
- Clarify that the alternative duration-factor coefficient forms may differ bitwise rather than always differing.

Tests:
- Strengthen bit-identity assertions, exact CAFEC/Z-index expectations, denominator error matching, and production-path coverage across all fitted divisions.

## Summary by CodeRabbit

* **Documentation**
  * Clarified documentation for duration-factor weighting and numerical precision considerations.

* **Tests**
  * Strengthened precision checks for duration-factor calculations and Palmer index results.
  * Updated validation coverage for fitted coefficients, recursion errors, and derivation forms.
  * Improved test coverage for handling invalid duration-factor inputs.


monocongo:
- **Review target:** [PR #969 — test(palmer): harden duration-factor and CAFEC bit-identity pins](https://github.com/monocongo/climate_indices/pull/969)
- **Lanes run:** 3 independent read-only lanes on `deepseek/deepseek-v4-pro` — tests, correctness/dataflow, maintainability. The tests lane was re-run on the fix delta as independent validation.
- **Validation:** `uv run ruff check src/ tests/` passed; `uv run ruff format --check src/ tests/` passed (89 files); `uv run mypy src/` passed (28 files); `uv run pytest` passed (2056 passed, 406 deselected); `uv run pytest -m validation` passed (378 passed, 3 skipped); focused Palmer suites passed (55 passed).
- **Fix commits:**
  - [2a41111b — test(palmer): name the division when a fitted coefficient is rejected](https://github.com/monocongo/climate_indices/commit/2a41111b7898873d6f0273a8729bf580c057e110)
  - [fb745cf1 — docs(palmer): drop the 1-ulp bound on coefficient-form divergence](https://github.com/monocongo/climate_indices/commit/fb745cf1797dbeb7adce41a8fb8cb7e77b4c5a9f)

### Findings and fix status

#### [P3] Docstrings bound coefficient-form divergence to one ulp, which is false — Fixed
- **Location:** `src/climate_indices/_palmer_duration.py` (lines 9–12, now line 10); `tests/test_palmer_wells.py` (line 155)
- **Origin:** Correctness/dataflow
- **Problem:** The PR softened "not bit-identical" to "may differ in the last bit of the float", but that still bounds the difference to 1 ulp. `DurationFactors.from_fitted(1e17, 1.0, 1e17, 1.0)` is accepted, and there `weighting_fraction` returns `1e-17` while the complement form returns `0.0` — the forms differ by the whole value, not one ulp. The softened wording was directionally right but left a false bound in place.
- **Applied fix:** State the difference as bitwise with an explicit note that it can exceed one ulp when the factors differ wildly in scale; test docstring now says "can differ bitwise". Commit `fb745cf1`.

#### [P3] Contraction test loses division attribution on hard failure — Fixed
- **Location:** `tests/test_scpdsi.py` (lines 242–251)
- **Origin:** Tests (independently raised by Maintainability)
- **Problem:** The PR replaced the test-local coefficient formulas with `DurationFactors.from_fitted`, which raises `ConvergenceError("...: dryc = 2.0 ...")` as soon as a coefficient reaches magnitude ≥ 1 — before the `worst` loop can record the division. The docstring still promises "erosion fails here, with the worst division named, rather than surfacing later as a ConvergenceError on user data". For the ≥ 1 band the test now surfaces exactly the unnamed ConvergenceError it says it prevents, a diagnostic regression versus the base inline computation.
- **Applied fix:** Wrap `from_fitted` in `try/except ConvergenceError` and `pytest.fail(f"{division_dir.name}: {error}")`, restoring the division name while keeping production-derivation coverage. Commit `2a41111b`.

#### [P4] Stale "CAFEC-weighting fraction" comment — Fixed
- **Location:** `tests/test_palmer_duration_factors.py` (lines 27–28)
- **Origin:** Maintainability
- **Problem:** The PR standardized the term as "duration-factor weighting fraction" in `_palmer_duration.py` and `CONTEXT.md`, but this comment (in a file the PR edits) still said "CAFEC-weighting fraction" — the only remaining use repo-wide, and it conflates the duration-factor fraction with the distinct K-factor weighting.
- **Applied fix:** Renamed to "duration-factor weighting fraction". Commit `fb745cf1`.

### Remaining blockers
None.

Issue #957, opened 2026-09-16
TDD_CLEANUP-13 — Migrate SPI parameter caching, then remove the legacy `spi` script


monocongo:
## Parent

#908 — follow-up from the retire decision in #919.

## What to build

The `spi` console script (`climate_indices.__spi__:main`) emits a deprecation warning as of 2.4.0 and is scheduled for removal in 3.0.0 (#919). Before removal:

1. Port `--save_params` / `--load_params` (NetCDF distribution-fitting parameter caching) into `climate_indices --index spi`, or record an explicit rationale for retiring the options.
2. Cover the migrated option(s) with behavior tests — fit once, reload the parameters, identical SPI output.
3. Delete:
   - `src/climate_indices/__spi__.py`
   - `tests/test_spi_console_script.py` (deprecation test added in #958)
   - the `spi` console-script entry point in `pyproject.toml`
   - the `[[tool.mypy.overrides]]` `ignore_errors` entry
4. Drop the remaining `__spi__` references from:
   - `docs/deprecations/api-changes.rst`, `docs/index.rst`, `docs/index.md`
   - `docs/architecture.md`, `docs/project-overview.md`, `docs/component-inventory.md`
   - `.agents/skills/climate_indices/SKILL.md`

   The three `docs/*.md` above are in `FULL_FILES` in `scripts/generate_llms_txt.py`, so run `uv run scripts/generate_llms_txt.py` and commit the regenerated bundles alongside the edits.

## Acceptance criteria

- [ ] Parameter caching has an equivalent in `climate_indices --index spi`, or a written rationale for dropping it.
- [ ] The equivalent behavior is covered by tests.
- [ ] Legacy module, entry point, mypy override, and its test file removed.
- [ ] No remaining `__spi__` references in source, tests, packaging, or docs.
- [ ] Full validation gate green.

## Blocked by

None for steps 1–2. Removal (steps 3–4) must not ship before the 3.0.0 deprecation window closes.

monocongo:
Decision on steps 1-2 (recorded by #962, still open for review): **the options are retired, not migrated.**

[...]

PR #1034, opened 2026-09-18
test(release): guard the 3.0.0 fire namespace in release checks (#1010)

### LINKED ISSUE #1010: RELEASE3-5 — Guard the 3.0.0 public surface in release integrity and wheel smoke tests
## What to build

3.0.0 adds the `fire` namespace to the public API. Two existing release checks do not cover it, so a broken export could ship:

1. `tests/test_release_integrity.py` — add a 3.0.0 public-surface test alongside `test_v240_public_api_importable` that imports `fire` from the installed package and asserts the fire entry points are callable.
2. `.github/workflows/release.yml` — the wheel-check job's smoke command imports only the drought API (`eddi`, `pci`, `percentage_of_normal`, `pet_hargreaves`, `pet_thornthwaite`, `spei`, `spi`). Extend it so the built wheel is proven to expose the fire namespace from outside the source checkout.

Both changes are guardrails: they must fail when the thing they guard is broken, not merely pass.

## Acceptance criteria

- [ ] The new test fails if `fire` is removed from the public `__init__` exports (verify by temporarily removing it, then restore)
- [ ] The wheel smoke command imports the fire namespace and still exercises all three CLI entry points
- [ ] `uv run pytest tests/test_release_integrity.py` passes, including the tests that parse `release.yml` for its guardrails
- [ ] ruff / format / mypy gates pass

## Blocked by

None - can start immediately

monocongo:
## Summary

- Adds `test_v300_public_api_importable` to `tests/test_release_integrity.py`: asserts `fire` is in `climate_indices.__all__`, is the same module object, and exposes callable `cffwis`, `fosberg_ffwi`, `haines_index`, `hot_dry_windy`, and `kbdi` entry points. Verified to fail when `fire` is removed from the public `__init__` exports, then restored.
- Extends the `wheel-check` smoke command in `.github/workflows/release.yml` to import `fire` from the installed wheel and assert the fire entry points are callable, with the matching workflow guardrail test updated.

## Scope

- The wheel-check entry-point loop already iterates every script declared in `pyproject.toml`; that is currently two (`climate_indices`, `process_climate_indices`) because the legacy `spi` console script was removed in #957. The issue's "all three CLI entry points" predates that removal; the loop stays derived from `pyproject.toml`, so any future third script is covered automatically.
- Only the `wheel-check` job's smoke command is extended; the `build` job's earlier smoke command stays as the pre-publish sanity check it already is.

## Test plan

- `uv run mypy src/` — pass
- `uv run pytest` — 2174 passed, 423 deselected
- `uv run pytest tests/test_release_integrity.py` — 33 passed
- Guard check: removing the `fire` export from `src/climate_indices/__init__.py` fails `test_v300_public_api_importable`; export restored afterward.

Closes #1010

## Summary by Sourcery

Protect the 3.0.0 fire public API with release integrity and installed-wheel checks.

Enhancements:
- Add release guardrails that verify the 3.0.0 fire namespace is publicly exported with its expected API and callable entry points.

CI:
- Extend the wheel-check workflow smoke test to validate the fire API in the installed package.

Tests:
- Add release integrity coverage for the public fire namespace and scope wheel smoke-test assertions to the wheel-check job.


monocongo:

- **Review target:** [PR #1034](https://github.com/monocongo/climate_indices/pull/1034) — `test/1010-release-surface-guard` → `main` (base `6c518fca`, head `b44fd8d4`)
- **Lanes run:** 3 read-only lanes — tests; public API contracts/packaging; maintainability
- **Validation:** `uv run ruff check src/ tests/` pass; `uv run ruff format --check src/ tests/` pass; `uv run mypy src/` pass; `uv run pytest` 2174 passed / 423 deselected; `uv run pytest tests/test_release_integrity.py` 33 passed. Guard removal re-checked: deleting only the eager `from climate_indices import fire` fails the fresh-interpreter assertion, and deleting it plus the `"fire"` entry from `__all__` fails the membership assertion; the export was restored afterward.
- **Fix commits:** [b44fd8d4 — test(release): pin fire.__all__ and scope the smoke guard to wheel-check](https://github.com/monocongo/climate_indices/commit/b44fd8d4aea249526ac68156cfbdab7e52c2ec33)

### Findings and fix status

#### [P2] Workflow guard could pass while the fire checks left the publishing gate — Fixed

- **Location:** `tests/test_release_integrity.py` (`test_release_workflow_smoke_tests_built_wheel`), `.github/workflows/release.yml` (build job vs `wheel-check` job)
- **Origin:** tests, contracts, maintainability lanes
- **Problem:** The new assertions matched the whole workflow file, and the nearly identical smoke command in the `build` job also satisfies the generic substrings. Moving or reverting the fire checks in `wheel-check` — the job that gates `publish` — would leave the test green.
- **Applied fix:** Extracted the `wheel-check` job section with the existing `split("\n  wheel-check:", ...)` idiom and asserted the smoke command against that section only. The `build` job's earlier command is deliberately left as its pre-publish sanity check, and the PR body documents that scope. Commit `b44fd8d4`.

#### [P2] In-process import masked a removed eager `__init__` export — Fixed

- **Location:** `tests/test_release_integrity.py` (`test_v300_public_api_importable`)
- **Origin:** tests, maintainability lanes
- **Problem:** `from climate_indices import fire` falls back to importing the submodule and rebinding the parent attribute, so the test still passed when `src/climate_indices/__init__.py` stopped importing `fire` while `__all__` still advertised it — the exact breakage the test's docstring promised to catch (`import climate_indices; climate_indices.fire` raising `AttributeError`).
- **Applied fix:** Added a fresh-interpreter check, `python -c "import climate_indices; assert climate_indices.fire"`, which cannot be satisfied by the fallback, plus the existing `"fire" in climate_indices.__all__` membership assertion. Commit `b44fd8d4`.

#### [P3] Only 5 of 24 `fire.__all__` names were guarded — Fixed

- **Location:** `tests/test_release_integrity.py` (`test_v300_public_api_importable`); `src/climate_indices/fire/__init__.py`
- **Origin:** contracts, tests lanes
- **Problem:** The five-name loop left `KBDIResult`, `KBDIState`, `haines_index_from_profile`, and other exports unpinned; a name added to `__all__` without a resolvable attribute would ship unnoticed.
- **Applied fix:** The test now asserts exact equality between `set(fire.__all__)` and the 3.0.0 surface, then asserts every name resolves to a callable. The wheel smoke keeps the five consumer-facing entry points explicit. Commit `b44fd8d4`.

#### [P3] Two independent five-name lists could drift apart — Fixed

- **Location:** `tests/test_release_integrity.py`, `.github/workflows/release.yml`
- **Origin:** contracts lane
- **Problem:** The test's tuple and the workflow's assertion list could be changed independently, so unit-level and wheel-level guards could silently cover different entry points.
- **Applied fix:** The test no longer carries its own tuple; it pins the exact workflow assertion string, making the workflow the single list and failing CI on any edit to it. Commit `b44fd8d4`.

#### [P3] PR body misstated how the CLI entry-point loop stays in sync — Fixed

- **Location:** PR #1034 description, "Scope" section
- **Origin:** maintainability lane
- **Problem:** The body claimed the loop is "derived from `pyproject.toml`"; it is a literal kept in sync by `test_release_workflow_installs_wheel_on_boundary_pythons`, which fails on any `pyproject.toml` script change until the workflow is updated.
- **Applied fix:** PR body corrected to describe the literal-plus-parity-test mechanism and the two declared scripts (`spi` was removed in #957). No code change.

### Remaining blockers

None. No P0/P1 findings; the two P2 findings are fixed and re-validated on `b44fd8d4`.

Issue #906, opened 2026-09-16
TIDY_UP-13: Test seams: shared Palmer fixture loaders, public duration-factor entry


monocongo:
## Files
- `tests/test_palmer.py:28-44`, `tests/test_scpdsi.py:22-33,154-160,238-261`, `tests/test_nclimdiv_reference.py:132-150` — repeated nClimDiv fixture loaders; calibration internals monkeypatched
- `tests/test_palmer_duration_factors.py:67,88,109,134,169,175-209` — calls `_statement_*` and replays `_initialize_data`→`_calc_*`→`_calc_zindex`→`_finish_up`
- `tests/test_palmer_wells.py:157-174` — constructs private `_Factors`/`_State` to pin an invariant
- `tests/test_cli_common.py:30-33`, `tests/test_main_kbdi.py:71,180,262` — fake xarray module, patched dispatch

## Problem
Tests enter through private types and hand-replayed pipelines because no public seam exists for (a) custom duration factors, (b) calibration diagnostics such as rescale-pass count. Private refactors therefore break tests without behavior changing, and the tests cannot prove the public path matches the internal one.

## Solution
- one `conftest.py` fixture loader for the nClimDiv division data;
- a supported way to pass duration-factor overrides into `pdsi` (or an explicitly documented prepared-pipeline seam) so `test_palmer_duration_factors` calls it;
- return calibration diagnostics (rescale passes, fitted coefficients) in the result params so `test_scpdsi.py:154-160` stops monkeypatching.

CLI dispatch seams improve as a side effect of TIDY_UP-7's registration work.

## Benefits
Tests pin behavior, not wiring; private types stay free to change.

## Recommendation strength
**Worth exploring**

## Related
#847, #851 (closed), TIDY_UP-6, TIDY_UP-7

## Acceptance
- Tests listed above no longer import private `_`-prefixed names where a public seam exists.
- `uv run ruff check src/ tests/`, `uv run ruff format --check src/ tests/`, `uv run mypy src/`, `uv run pytest` pass.

monocongo:
Closing as delivered. #1053 added the public seams the ticket asked for: `pdsi()` honors `wetm`/`wetb`/`drym`/`dryb` in `fitting_params` as a validated duration-factor override, `scpdsi()` reports `rescale_passes` in its returned params, and `conftest.py` now owns the shared nClimDiv division fixture loaders. #1057 fixed the override's validation order.

Deferred, as recorded on the PR:
[...]

Issue #902, opened 2026-09-16
TIDY_UP-09: xarray validation facade: stop reaching into adapter privates


monocongo:
## Files
- `src/climate_indices/fire.py` — private adapter imports (:67-73), `_kbdi_xarray` (:948-1213), `_hdw_xarray` (:2393-2491)
- `src/climate_indices/xarray_adapter.py` — validators (:562, :592, :1278), `_DailyCalendarPlan` (:290-462), decorator (:1497)
- `src/climate_indices/_cli.py:17-77`, `src/climate_indices/__main__.py:134-191,627-694,666-675`
- Three `InputType` concepts: `__main__.py:37`, `xarray_adapter.py:80`, `__spi__.py:36-45`

## Problem
The adapter's private names are the de facto interface: `fire.py:67-73` imports four `_`-prefixed helpers, and `_kbdi_xarray` hand-rolls alignment/units/attr logic the decorator already owns. Time/dimension validation is re-implemented in `_cli.py`, `__main__.py`, the adapter, and KBDI's xarray path. Daily calendar normalization is duplicated (`__main__.py:666-675` vs the adapter's `_DailyCalendarPlan`), and KBDI bypasses the shared CLI path because of it (`__main__.py:1624-1632`). Three same-named `InputType` enums make detection results non-interchangeable.

## Solution
A public validation façade (e.g. `climate_indices.validation`) owning time/dimension checks, calendar planning, and input-kind detection; adapter, CLI, and fire all call it. Promote or rename the four adapter helpers it exports. One `InputType`. Then KBDI's xarray wrapper can use the decorator instead of its bespoke path.

## Benefits
Leverage: one validation contract for every entry surface. Locality: KBDI stops carrying its own copy. Testability: validators testable without a fake xarray module (`tests/test_cli_common.py:30-33`).

## Recommendation strength
**Worth exploring**

## Related
#759 (`_resolve_periodicity`), #661 (CLI metadata), #834 (calendar/docs), #885

## Acceptance
- Validation behavior unchanged; KBDI xarray path still passes its fixtures.
- `uv run ruff check src/ tests/`, `uv run ruff format --check src/ tests/`, `uv run mypy src/`, `uv run pytest` pass.

monocongo:
## First slice in review (#1021)

`refactor(validation): add a public validation facade for shared input checks`

The unblocked slice of this issue ships in #1021:

- `climate_indices.validation` now owns `InputType`, `detect_input_type`,
[...]

PR #1053, opened 2026-09-18
refactor(palmer): add public duration-factor and calibration-diagnostic seams (#906)

monocongo:
## Summary

TIDY_UP-13's test seams: the tests that replayed private Palmer internals now
enter through public behavior, and the duplicated division-fixture loading
lives in one place.

- `pdsi()` now honors all four of `wetm`/`wetb`/`drym`/`dryb` in
  `fitting_params` as a validated duration-factor override (finite scalars,
  validated through `DurationFactors.from_fitted`, all-or-nothing). This is
  the supported entry the ticket asked for; `test_custom_duration_factors_change_pdsi_output`
  now calls `pdsi()` twice instead of rebuilding the preparation pipeline
  through `_initialize_prepared`/`_calc_*`/`_initialize_recursion`.
- `scpdsi()` reports `rescale_passes` (the fixed 3) in its returned parameter
  dictionary, so the rescaling loop count is asserted as public output
  instead of by monkeypatching `_palmer_wells.calculate`. The cumulative
  behavior those passes produce stays pinned end-to-end by
  `test_scpdsi_calibration_anchor_lands_on_target` across all 344 divisions.
- `conftest.py` now owns the division-fixture loading:
  `palmer_division_dirs`, the parametrized `palmer_division_dir`, and
  `palmer_division_inputs` (division-keyed `(precips, pet, awc)`, session
  scoped). The sweeps, `test_scpdsi`, `test_nclimdiv_reference`, and
  `test_palmer` all read the same loader; the local `division_dirs` fixture
  and `_division_inputs`/`_AWCS` helpers are gone. `test_palmer_sweep_fixture`
  still pins the per-division failure isolation, now against the input map.

A side effect of sharing the session-scoped inputs: the infinity test now
copies before mutating, since it had been quietly writing `inf` into the
100-year series that later tests reuse.

## Scope decisions

- **Duration-factor override rides `fitting_params`, not a new keyword.**
  `test_public_signature_matches_pdsi_and_is_exported` pins `scpdsi()`'s
  signature to `pdsi()`'s except `spatial_time_major`, and scPDSI's
  documented contract is that duration factors are always recalibrated.
  A new pdsi-only kwarg would break that parity for a capability scPDSI
  cannot honor anyway. Supplying the four keys to `scpdsi()` keeps its
  existing recalibrate-and-ignore behavior, now stated in its docstring.
- **Deferred: promoting the prepared pipeline (`_initialize_prepared`,
  `_calc_*`, `_statement_*`) to public API.** The statement-level duration-
  factor tests still import private names because no public seam exists for
  that arithmetic, and exposing the whole stage structs is an API commitment
  larger than this ticket. That promotion belongs with TIDY_UP-6 (#899),
  which remains deliberately open for it.
- **Deferred: CLI and xarray fake-module test seams**
  (`test_cli_common.py`, `test_main_kbdi.py`) to TIDY_UP-9 (#902), whose
  validation façade is the seam those tests need; the ticket's own text
  routes them there.

## Validation

- `uv run mypy src/`
- `uv run pytest` — 2199 passed
- `uv run pytest -m validation tests/test_scpdsi.py tests/test_nclimdiv_reference.py tests/test_palmer.py`
  — 349 passed, including the 344-division bit-identical scPDSI oracle and
  the nClimDiv comparisons.

## Summary by Sourcery

Expose supported Palmer calibration diagnostics and duration-factor controls through the public APIs while centralizing division-fixture reuse and validating behavior end to end.

New Features:
- Allow standard PDSI callers to provide validated wet and dry duration-factor overrides through fitting parameters.
- Expose effective duration factors in PDSI results and the fixed scPDSI rescaling-pass count in returned diagnostics.

Bug Fixes:
- Prevent shared Palmer test inputs from being mutated by tests, avoiding cross-test contamination.

Enhancements:
- Consolidate climate-division fixture discovery and input loading into shared session-scoped fixtures while preserving per-division failure isolation.
- Replace tests that exercised private Palmer internals or monkeypatched recursion behavior with public-contract and end-to-end assertions.

Documentation:
- Document duration-factor override behavior, returned diagnostic parameters, and scPDSI's recalibration semantics.

Tests:
- Add coverage for duration-factor overrides, round-trip parameter reuse, validation failures, wet/dry distinction, scPDSI rescaling diagnostics, and cumulative rescaling behavior.


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/1053
- **Lanes run:** 4 first pass (correctness/edge cases, public API contracts, tests/coverage, maintainability) + 2 second pass on the fix commits (correctness, tests)
- **Validation:** `uv run ruff check src/ tests/` pass; `uv run ruff format --check src/ tests/` pass; `uv run mypy src/` pass; `uv run pytest` 2220 passed, 428 deselected; `uv run pytest -m validation tests/test_scpdsi.py tests/test_nclimdiv_reference.py tests/test_palmer.py` 349 passed, 26 deselected
- **Fix commits:**
  - [1666275b — fix(palmer): harden the duration-factor override and keep fixture loads isolated (#906)](https://github.com/monocongo/climate_indices/commit/1666275b)
  - [a34fdd53 — style(tests): drop a trailing blank line in the override test (#906)](https://github.com/monocongo/climate_indices/commit/a34fdd53)
  - [52c1b42e — test(palmer): pin distinct duration factors and non-vacuous rescaling (#906)](https://github.com/monocongo/climate_indices/commit/52c1b42e)

### Findings and fix status

#### [P2] Shared fixture loader aborted the session on one bad division — Fixed
- **Location:** `tests/conftest.py` (`palmer_division_inputs`, `_palmer_sweep`)
- **Origin:** Correctness, Maintainability
- **Problem:** Moving the division loads into the session fixture put `np.load` and the `palmer_awc.json` lookup outside any guard, so one unreadable division would error the whole fixture — contradicting the sweeps' documented per-division isolation.
- **Applied fix:** `palmer_division_inputs` records a per-division `RuntimeError` in a `_PalmerSweep` when a load fails, and `_palmer_sweep` indexes its inputs inside the existing `try` so a stored load failure is isolated per division. Arrays are set read-only so a future in-place test mutation cannot leak between consumers. Pinned by two new tests in `tests/test_palmer_sweep_fixture.py`. Commit `1666275b`.

#### [P2] `pdsi()`'s returned params dropped the duration-factor override — Fixed
- **Location:** `src/climate_indices/palmer.py` (`_calculate_pdsi_prepared`, `pdsi` docstring)
- **Origin:** Correctness, Contracts
- **Problem:** The override was applied to the run but not echoed in the returned parameter dictionary, so the natural save-and-reuse flow (`pdsi(..., fitting_params=result_params)`) silently fell back to Palmer's national constants.
- **Applied fix:** the returned dict now includes the effective `wetm`/`wetb`/`drym`/`dryb` (with the spatial-block re-fit caveat documented), and `test_pdsi_returned_params_reproduce_a_duration_factor_override` asserts the echoed values and exact rerun equality. Commits `1666275b`, `52c1b42e`.

#### [P2] Cross-index reuse: a scPDSI params dict passed to `pdsi()` now applies its duration factors — Fixed (deliberate, documented)
- **Location:** `src/climate_indices/palmer.py` (`_duration_factor_override`, `pdsi`/`scpdsi` docstrings)
- **Origin:** Contracts
- **Problem:** Pre-branch, `wetm`/`wetb`/`drym`/`dryb` in `fitting_params` were ignored by `pdsi()`. They are now honored, which changes behavior for callers who reused a saved scPDSI parameter dict that still carries those keys.
- **Applied fix:** this is the ticket's requested override and is now documented in both docstrings; the round-trip fix above makes the new behavior lossless in the intended direction. The mechanism stays on `fitting_params` rather than a new keyword because `test_public_signature_matches_pdsi_and_is_exported` pins pdsi/scpdsi signature parity except `spatial_time_major`, and scPDSI cannot honor a duration-factor override. Not noted in `CHANGELOG.md`: this repo maintains it at release time, and recent merged refactor PRs do not touch it.

#### [P2] Default test suite lost its deterministic pin of the rescaling loop — Fixed
- **Location:** `tests/test_scpdsi.py`
- **Origin:** Tests, Maintainability
- **Problem:** `rescale_passes == 3` echoes the loop bound, so a 1-pass, 2-pass, or non-cumulative rewrite could pass it; the deleted monkeypatch test was the only deterministic pin in the default (non-`validation`) suite.
- **Applied fix:** `test_scpdsi_rescales_zindex_cumulatively` pins constant percentile anchors through the public `self_calibration` module and asserts the returned scPDSI Z-index equals the raw Z times exactly 2³; the second pass added an `np.isfinite(raw_z).any()` guard so an all-NaN Z cannot pass vacuously. Commits `1666275b`, `52c1b42e`.

#### [P3] `OverflowError` escaped the documented `ValueError` — Fixed
- **Location:** `src/climate_indices/palmer.py` (`_duration_factor_override`)
- **Origin:** Correctness
- **Problem:** `fitting_params={"wetm": 10**1000, ...}` raised `OverflowError` even though the docstring promises `ValueError` for a non-finite scalar.
- **Applied fix:** `OverflowError` added to the conversion catch; a parametrized `overflow` case pins it. Commit `1666275b`.

#### [P3] A pdsi override's `ConvergenceError` named the Wells/scPDSI calibration — Fixed
- **Location:** `src/climate_indices/palmer.py` (`_duration_factor_override`)
- **Origin:** Contracts
- **Problem:** `DurationFactors.from_fitted` is scPDSI-framed, so a non-contracting override raised a message about the Wells recursion with `algorithm="scPDSI duration-factor calibration"` from `pdsi()`.
- **Applied fix:** the error is re-attributed with a pdsi-specific message and `algorithm="PDSI duration-factor override"`, preserving the original as `__cause__`/`underlying_error`; pinned by `test_non_contracting_duration_factor_override_is_attributed_to_pdsi`. Commit `1666275b`.

#### [P3] Malformed override raised before any lifecycle logging — Fixed
- **Location:** `src/climate_indices/palmer.py` (`_palmer_calculation`, `pdsi`)
- **Origin:** Correctness, Maintainability
- **Problem:** `_duration_factor_override` was evaluated in `pdsi()`'s argument list, so its errors emitted no `calculation_started`/`calculation_failed` events, unlike other input-validation failures.
- **Applied fix:** pdsi now passes `use_fitting_duration_factors=True` and the override is resolved at the top of `_palmer_calculation`'s `try`, inside the bound logger's lifecycle. scPDSI still never parses those keys. Commit `1666275b`.

#### [P4] Custom-override test could pass on an all-NaN difference — Fixed
- **Location:** `tests/test_palmer_duration_factors.py`
- **Origin:** Tests
- **Problem:** `assert not np.allclose(default, custom, equal_nan=True)` also passes when the custom run became all-NaN.
- **Applied fix:** added a finite-values assertion and an equal-NaN-mask assertion before the inequality. Commit `1666275b`.

#### [P4] Override validation branches under-tested — Fixed
- **Location:** `tests/test_palmer_duration_factors.py`
- **Origin:** Tests
- **Problem:** Only the missing-key and list-valued branches were exercised, with one shared match string.
- **Applied fix:** parametrized partial / non-scalar / non-finite / non-numeric / overflow cases with per-branch match strings. Commit `1666275b`.

#### [P4] Stale and contradictory internal docs — Fixed
- **Location:** `src/climate_indices/palmer.py` (`_select_duration_factors`, `_initialize_prepared` override comment, `pdsi` Raises/Returns)
- **Origin:** Contracts, Maintainability
- **Problem:** `_select_duration_factors` still claimed pdsi always uses the identical default pair; the new override comment said the duration factors are "shared with scPDSI"; the spatial-block round-trip claim was inaccurate.
- **Applied fix:** all three corrected in `1666275b`/`52c1b42e`, and `pdsi`'s `Raises` section now covers the pre-existing shared `ValueError` cases too.

#### [P4] Round-trip and rescale tests used all-equal values (second pass) — Fixed
- **Location:** `tests/test_palmer_duration_factors.py`, `tests/test_scpdsi.py`
- **Origin:** Correctness, Tests (second pass)
- **Problem:** With `{1,1,1,1}` a transposed wet/dry echo still reproduced the run, and no test distinguished the wet and dry pairs end-to-end.
- **Applied fix:** distinct factors `{1, 2, 3, 4}`, an explicit echoed-key assertion, and a wet/dry swap inequality test; all-NaN guard added to the rescale pin. Commit `52c1b42e`.

### Remaining blockers
None. No unresolved P0/P1; all feasible findings were fixed and the full gate plus the validation suite pass on the final HEAD.

PR #1057, opened 2026-09-18
fix(palmer): resolve the duration-factor override after input validation (#906)

monocongo:
## Summary

PR #1053 merged the public duration-factor seam. A follow-up multi-agent review pass over the merged code found four residual defects/drift items that the pre-merge hardening did not cover. All are fixed here, on top of current `main` (`196046cf`).

- **The override was resolved before input validation.** `use_fitting_duration_factors` resolved it at the top of `_palmer_calculation`'s `try`, i.e. ahead of the shape/infinite checks, the all-missing fast path, and `_prepare_palmer_data`'s calibration-period validation. Consequences on `main`: all-NaN input plus a partial override raised instead of returning the documented four NaN arrays and `None`; infinite input plus a bad override reported the override error, hiding the input error the caller can act on; a bad calibration period likewise. It is now resolved once `_prepare_palmer_data` returns and applied to the prepared fields there — inside the logged block (so a malformed override still emits `calculation_started`/`calculation_failed`), after every input-validation stage, and before the recursion, the only reader. This removes the `duration_factors` argument from `_prepare_palmer_data`.
- **Masked override values were accepted as their backing data.** `np.asarray` drops a mask, so `np.ma.masked` reached the recursion as `0.0` and surfaced later as a misleading `ConvergenceError` instead of the documented `ValueError`. Masked values are now rejected at the gate, matching the module's own `_fill_masked_with_nan` stance.
- **Default-path equivalence was untested.** Nothing pinned that passing Palmer's own constants through the override reproduces the run that never used them bit-for-bit. A test now does, alongside the existing wet/dry and round-trip pins.
- **Documentation drift.** `_PalmerPrepared` was described as read-only after preparation (the override writes its four fields); `CONTEXT.md` and `DurationFactors.from_defaults` still said standard PDSI uses the national constants unconditionally; the `pdsi()` trailing-shape clause and `_palmer_cafec_params` docstring said "when `fitting_params` was supplied" where only a complete, valid CAFEC coefficient set changes the shape; `pdsi()`'s `Raises` gained the ambiguous-block and non-broadcastable-`awc` conditions plus the all-missing exemption; `scpdsi()`'s `Raises` gained the infinite-value and calibration-period conditions it shares with the same code path; and the override is now documented as producing neither Palmer's (1965) standard index nor scPDSI.

## Tests

- masked case added to the override-rejection parametrization
- override ordering pinned against infinite values and a bad calibration period
- all-missing return (`None` params, NaN arrays) pinned against a partial override
- Palmer's defaults supplied through the override must reproduce the default run bit-for-bit
- unmarked 0101 `sczindex` oracle comparison, pinning the rescaling result outside the `validation` marker
- redundant `division_dir = palmer_division_dir` alias dropped

## Validation

- `uv run mypy src/`
- `uv run pytest` — 2239 passed, 428 deselected
- `uv run pytest -m validation tests/test_scpdsi.py tests/test_nclimdiv_reference.py tests/test_palmer.py` — 349 passed

## Summary by Sourcery

Correct Palmer duration-factor override handling so input validation takes precedence and malformed or masked overrides are rejected consistently.

Bug Fixes:
- Resolve duration-factor overrides only after Palmer input preparation and validation, preserving the correct errors and all-missing behavior.
- Reject masked duration-factor override values instead of treating their backing data as valid factors.

Enhancements:
- Document duration-factor overrides, CAFEC parameter shape behavior, and shared validation contracts for PDSI and scPDSI.

Documentation:
- Clarify that standard PDSI duration factors can be overridden and that overridden results differ from Palmer's standard index and scPDSI.

Tests:
- Add coverage for override validation ordering, masked values, all-missing inputs, and bit-for-bit equivalence with Palmer's default factors.
- Pin an unmarked scPDSI Z-index oracle result.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of custom PDSI duration factors, ensuring each supplied value is applied to the correct calculation.
  - Invalid, masked, or malformed duration-factor values now produce clear validation errors.
  - Input validation continues to take precedence, while all-missing inputs retain their existing behavior.
  - Preserved scPDSI calculation results and compatibility with established reference values.

- **Documentation**
  - Clarified how standard PDSI duration-factor defaults can be overridden and how scPDSI uses location-specific factors.


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/1057 (`fix/906-override-validation-order`, reviewed `293c03fb`, fixed HEAD `408f1d84`, base `196046cf`)
- **Lanes run:** 11 dispatch attempts, 7 successful lane results. First pass: correctness/edge cases, tests, security/hostile-input, docs/contracts, and numerical methodology/dataflow; the docs and methodology lanes were retried on deepseek after an openai-codex usage limit and Google free-tier quota exhaustion (both provider failures recorded, no lane claimed unavailable without an attempt). Post-fix: correctness and tests verification lanes on the fixed HEAD.
- **Validation (on `408f1d84`):** `uv run ruff check src/ tests/` pass; `uv run ruff format --check src/ tests/` pass; `uv run mypy src/` pass; `uv run pytest` 2241 passed, 428 deselected; `uv run pytest -m validation tests/test_scpdsi.py tests/test_nclimdiv_reference.py tests/test_palmer.py` 349 passed.
- **Fix commits:** [`e3f1d97a` — fix(palmer): treat masked fitting parameters as missing and correct override docs](https://github.com/monocongo/climate_indices/commit/e3f1d97a) · [`408f1d84` — fix(palmer): tighten scpdsi Raises wording and pin the mask-attribute gate](https://github.com/monocongo/climate_indices/commit/408f1d84)

### Findings and fix status
#### [P3] Denormal duration-factor override silently overflows the recursion to `inf`/`NaN` — Blocked (pre-existing, needs a methodology decision)
- **Location:** `src/climate_indices/_palmer_duration.py` (lines 83–96), `src/climate_indices/palmer.py` (lines 1379–1384)
- **Origin:** Security/hostile-input lane
- **Problem:** `pdsi(..., fitting_params={"wetm": 5e-324, "wetb": 5e-324, "drym": 5e-324, "dryb": 5e-324})` passes the finite-scalar and contraction checks, then `state.z / (wetm + wetb)` overflows: reproduced 48/48 non-finite output values with only a numpy `RuntimeWarning`. This is unchanged from `main` (`DurationFactors.from_fitted` is identical there), but it is the same gate this PR hardens.
- **Applied fix:** **Blocked.** Rejecting it requires choosing a minimum denominator (`z / (m + b)` can overflow for any small-but-normal denominator given a large enough `z`), i.e. a methodology decision about which factors are numerically meaningful, not a mechanical guard. The lane's suggested floor (1e-8, against real calibrated denominators of O(1)) is a reasonable starting point. Follow-up: decide and enforce the floor in `DurationFactors.from_fitted`, which silently converts today's non-finite output into the documented `ConvergenceError`.

#### [P4] Masked CAFEC coefficients were adopted as their backing values — Fixed
- **Location:** `src/climate_indices/palmer.py` (lines 1132–1148), covered by `tests/test_palmer_data.py` (line 67)
- **Origin:** Methodology/dataflow and security lanes
- **Problem:** `np.asarray` drops a mask, so a caller-supplied CAFEC coefficient with a masked element was accepted as its backing data: `alpha = np.ma.array(..., mask=[True, False, ...])` produced `prepared.calibrate is False` and used the masked value. The sibling duration-factor gate already rejected masks; this path did not.
- **Applied fix:** Treat a masked coefficient as missing data and fall back to fitting like any other malformed set; the parametrized malformed-fitting-params test gained a `masked` case. Commit `e3f1d97a`.

#### [P4] `np.ma.is_masked` on a non-masked object with a bare `_mask` attribute raised `AttributeError` — Fixed
- **Location:** `src/climate_indices/palmer.py` (lines 1370–1382), covered by `tests/test_palmer_duration_factors.py` (lines 302–308)
- **Origin:** Post-fix correctness lane
- **Problem:** `np.ma.is_masked` reads a `_mask` attribute off any object and calls `.any()` on it, so an object with `_mask = False` (or a string) escaped the override gate as `AttributeError` instead of the documented `ValueError`.
- **Applied fix:** Type-check with `np.ma.isMaskedArray` before `np.ma.is_masked`, and pin the bare-`_mask` case in the rejection parametrization. Commits `e3f1d97a`, `408f1d84`.

#### [P4] `pdsi()` Args claimed CAFEC reuse for a partial coefficient set — Fixed
- **Location:** `src/climate_indices/palmer.py` (lines 1727–1731)
- **Origin:** Docs/contracts lane
- **Problem:** "the CAFEC `alpha`/`beta`/`gamma`/`delta` coefficients stay fitted from the record unless supplied alongside" contradicted `_validate_fitting_params`, which discards an incomplete set and re-fits all four.
- **Applied fix:** Wording now says "unless a complete, valid set is supplied alongside (a partial set is discarded and all four are re-fitted)", matching the Returns clause. Commit `e3f1d97a`.

#### [P4] `scpdsi()` duration-factor fit/rescale order was documented in reverse — Fixed
- **Location:** `src/climate_indices/palmer.py` (lines 1731–1736)
- **Origin:** Docs/contracts lane
- **Problem:** The new sentence said scPDSI calibrates factors against "its own K-prime and rescaled Z series"; `_calculate_scpdsi_prepared` fits the factors from the K-prime Z series, then rescales that series around the now-fixed factors.
- **Applied fix:** Sentence reordered to K-prime fit then rescale, matching `VALIDATION.md`. Commit `e3f1d97a`.

#### [P4] `Raises` clauses omitted the 2-D invalid-second-dimension `ValueError` — Fixed
- **Location:** `src/climate_indices/palmer.py` (lines 1757–1762, 1827–1832)
- **Origin:** Docs/contracts lane; post-fix correctness lane caught the scpdsi wording
- **Problem:** Both `pdsi()` and `scpdsi()` raise `ValueError: Values array has an invalid shape (2-D but second dimension not 12)` for input such as `(2, 13)`, but neither `Raises` clause named it.
- **Applied fix:** Both clauses now name the 2-D second-dimension condition; the scpdsi clause says "different numbers of elements" rather than "shapes", because a 1-D series against a same-size `(years, 12)` array is deliberately accepted. Commits `e3f1d97a`, `408f1d84`.

#### [P4] Defaults-as-override test overclaimed slot routing — Fixed
- **Location:** `tests/test_palmer_duration_factors.py` (lines 326–343)
- **Origin:** Tests lane
- **Problem:** The test named and documented itself as pinning "each slope/intercept slot", but `DurationFactors.from_defaults()` has `wetm == drym` and `wetb == dryb`, so a wet/dry swap or a dropped mapping would still pass. Key-to-field routing is in fact pinned by `test_pdsi_returned_params_reproduce_a_duration_factor_override` with distinct `1/2/3/4` values.
- **Applied fix:** Renamed to `test_palmer_default_factors_as_override_reproduce_the_default_run` and the docstring now states only the bit-for-bit equivalence it actually pins, pointing at the routing test. Commit `e3f1d97a`.

#### [P4] Override-ordering test omitted the shape-error case — Fixed
- **Location:** `tests/test_palmer_duration_factors.py` (line 358)
- **Origin:** Tests lane
- **Problem:** The ordering regression test pinned infinite-value and calibration-period precedence but not the shape check, which on `main` was also preempted by a partial override.
- **Applied fix:** Added an `Incompatible precipitation and PET arrays` assertion to the same test. Commit `e3f1d97a`.

#### [P4] Masked-override coverage did not exercise the new gate branch — Fixed
- **Location:** `tests/test_palmer_duration_factors.py` (lines 296–308)
- **Origin:** Tests lane (both passes)
- **Problem:** The initial masked parametrization covered `np.ma.masked`; the 0-d masked-array form and (after the gate change) the non-masked bare-`_mask` object were unpinned.
- **Applied fix:** Added `masked-array` (`np.ma.array(1.0, mask=True)`) and `bare-mask-attribute` (`types.SimpleNamespace(_mask="x")`) parameters. Commits `e3f1d97a`, `408f1d84`.

Rejected as non-findings (validated, intentional or contrived, recorded for audit): the all-missing fast path deliberately skips override validation (documented in `pdsi()` and pinned by `test_duration_factor_override_is_not_validated_for_all_missing_input`); the P3 denormal item above is the only unfixed finding.

### Remaining blockers
None — no P0/P1 findings. The P3 denormal-override item is pre-existing on `main` and awaits a methodology decision on a minimum denominator before it can be fixed; the rest are fixed at HEAD `408f1d84` (two commits pushed on top of the reviewed head `293c03fb`).

Issue #995, opened 2026-09-17
DOCS-9: Write the missing how-to guides


monocongo:
## Parent

Part of DOCS-0 (#986)

## What to build

Write the how-to guides the audit found missing, as task-shaped recipes under the How-to section: Dask-backed gridded computation and time-chunk constraints; parallel and cluster execution; writing outputs (NetCDF and Zarr); choosing a distribution and calibration period; input data requirements; and performance expectations with benchmark guidance. Each guide states the goal, prerequisites, steps, and how the reader verifies success, teaches nothing, and cross-links to the relevant reference pages and notebooks. Existing open documentation requests in this area are satisfied or explicitly deferred on their own issues.

## Acceptance criteria

- [ ] A how-to guide exists for each listed job and is reachable from the How-to section
- [ ] Each guide is task-shaped: goal, prerequisites, steps, verification; no tutorial-style teaching
- [ ] Guides cross-link to relevant reference pages and example notebooks
- [ ] Examples match current CLI and API behavior and are covered by the DOCS-1 gate where executable
- [ ] Existing open documentation requests for input data requirements and performance guidance are updated (closed or explicitly deferred)
- [ ] The DOCS-1 gate is green

## Blocked by

- DOCS-7 (##992)

PR #1079, opened 2026-09-19
docs(how-to): correct claims found in review (#995)

monocongo:
Follows #1076 (merged), which delivered #995. The multi-agent review of that PR ran after it was merged; this PR carries the corrections it found. No behavior changes: documentation only, plus the regenerated `llms-full.txt`.

## Corrections

- **In-memory xarray inputs with missing values raise, they do not only warn.** `docs/choosing-parameters.md` said a short calibration period emits `ShortCalibrationWarning` and an out-of-range request is silently widened. For an in-memory xarray input containing NaNs, `src/climate_indices/xarray_adapter.py` raises `InsufficientDataError` when the window holds fewer than 30 effective non-NaN years, and for a window with no overlap. The page now separates the NumPy warning path from the xarray check.
- **Result metadata records the requested calibration years, not the years a widened fit used.** `compute.py` adjusts the bounds for its own fit without writing them back, so a request outside the record leaves `calibration_year_initial`/`_final` describing a period the fit did not use. The page now says so and drops the claim that the metadata always names the actual baseline.
- **SPEI's Pearson failure is described by policy, not by one exception type.** `fallback_to_gamma=False` for SPEI, but an insufficient-non-zero fit raises `InsufficientDataError` while a failed L-moment fit raises `PearsonFittingError`.
- **Reused gamma parameters do not fix the zero mass.** `compute.py` recomputes `probabilities_of_zero` from the values each call receives, so `fitting_params` fixes the continuous part only.
- **Per-cell `fitting_params` do not align on Dask.** The xarray adapter passes one `fitting_params` value to every block, so gridded Dask reuse needs period-only parameters (or no reuse); the page now scopes the advice.
- **Command-line file naming is scoped to SPI and SPEI.** PNP writes `..._pnp_03.nc`, PET writes `..._pet_thornthwaite.nc`, and the Palmer family uses its own variable names.
- **CLI provenance limits stated.** A command-line NetCDF records `long_name` and the valid range only; the distribution and scale live in the file name, so the run arguments have to be kept alongside it.
- **Smaller fixes:** the `zarr` package is named as a prerequisite (it is not a runtime dependency), the Zarr example states that it creates a store rather than replacing one, the `--chunksizes` behavior claim is replaced with a pointer to the CLI reference, one mis-attributed recommendation is gone, and the performance page now places the notebook client's teardown in the write cell that actually closes it.

## Deferred finding (not fixed here)

`--chunksizes input` is a silent no-op with the inputs the CLI opens by default. `_input_chunksizes` in `src/climate_indices/__main__.py` gates on `da.encoding.get("contiguous", True)`, and the h5netcdf backend that xarray selects here does not set `contiguous`, so the guard skips every variable and the output is written contiguous. Reproduced with `--index pnp --chunksizes input` over a chunked h5netcdf input: the output file reports `chunksizes: None`. `docs/reference.md` documents the copy, so either the gate wants `False` as its default (`chunksizes` being present is the real signal) or the reference needs the caveat. This is a pre-existing CLI behavior change rather than a documentation defect, so it is left out of a docs PR; filed as #1080.

## Verification

- `sphinx-build -E -b html -W --keep-going docs docs/_build/html` — green
- `sphinx-build -E -b doctest docs docs/_build/doctest` — 76 tests, 0 failures
- `uv run pytest` — 2259 passed, 432 deselected, 0 failures
- `ruff check src/ tests/`, `ruff format --check src/ tests/`, `mypy src/` — clean
- Every fenced Python example in both pages was extracted and executed against this revision
- `uv run scripts/generate_llms_txt.py` — no diff beyond the committed regeneration


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/1079 (corrections follow-up to #1076, which delivered #995; the review ran against #1076's diff at `ee3bd8d4` and the fixes are committed here)
- **Lanes run:** 4 read-only sub-agent lanes — documentation factual accuracy (openai-codex), correctness and edge cases (deepseek), public API contracts and metadata (deepseek), tests and verification adequacy (deepseek)
- **Validation:** `sphinx-build -E -b html -W --keep-going` green; `sphinx-build -E -b doctest` 76 tests, 0 failures; `ruff check`, `ruff format --check`, `mypy src/` clean; every fenced Python example in the two new pages extracted and executed; `llms-full.txt` regenerated and byte-matches the generator; full `pytest` suite green on the corrected HEAD (2259 passed, 432 deselected, 0 failures)
- **Fix commits:** [`cafcb518` — docs(how-to): correct claims found in review (#995)](https://github.com/monocongo/climate_indices/commit/cafcb518)

### Findings and fix status

#### [P2] The xarray calibration-window contract was documented as warnings-only — Fixed
- **Location:** `docs/choosing-parameters.md` (calibration bullets, now lines 47–62)
- **Origin:** Documentation accuracy, correctness, contracts (three lanes, deduplicated)
- **Problem:** The page said a short period warns and an out-of-range request is silently widened. For an in-memory xarray input containing NaNs, `xarray_adapter.py` raises `InsufficientDataError` when the window holds fewer than 30 effective non-NaN years or has no overlap; the NumPy warning path is the other case. A reader following the guide's own example (marked missing values) hits an error the page never mentioned.
- **Applied fix:** The page now separates the NumPy warning path from the in-memory xarray check, names `InsufficientDataError`, states that Dask-backed input skips the check, and defers the full contract to `data_requirements.md`. Commit `cafcb518`.

#### [P2] Result metadata records the requested calibration years, not the fitted ones — Fixed
- **Location:** `docs/choosing-parameters.md` (calibration bullets and closing paragraph)
- **Origin:** Documentation accuracy, correctness
- **Problem:** `compute.py` widens out-of-range bounds only for its own fit, while `xarray_adapter.py` writes the caller's values into the result attributes. Verified by execution: a request of 1900–1905 over 1981–2020 data returns `calibration_year_initial=1900`, `calibration_year_final=1905`. The page claimed the metadata states the baseline that was standardized against, and advised passing years explicitly for that reason.
- **Applied fix:** The page now says the recorded years are the requested ones, tells the reader to keep the request inside the coverage when the two must agree, and adds a coverage print to the verification snippet. Commit `cafcb518`.

#### [P2] Per-cell `fitting_params` do not align per block on Dask — Fixed (documented; library limitation tracked)
- **Location:** `docs/choosing-parameters.md` (fit-reuse paragraph)
- **Origin:** API contracts
- **Problem:** The adapter passes one `fitting_params` value to every block (`xarray_adapter.py` `valid_kwargs` → per-block wrapper; `fitting_params` is never an `apply_ufunc` input). A lane reproduced both a broadcast `DistributionFittingError` and, when the shapes happen to coincide, silently wrong SPI values (2 of 8 cells, max |Δ| = 1.21). The page presented reuse as unconditionally valid.
- **Applied fix:** The page now scopes reuse to the NumPy API and whole-grid in-memory calls, and recommends period-only parameters for Dask. Commit `cafcb518`. The adapter-side fix (carrying cell dimensions through `apply_ufunc`) is a separate change, recorded below.

#### [P2] Reused gamma parameters do not fix the zero mass — Fixed
- **Location:** `docs/choosing-parameters.md` (fit-reuse paragraph)
- **Origin:** API contracts
- **Problem:** `compute.py` recomputes `probabilities_of_zero = zeros / values.shape[0]` from the values each call receives, regardless of supplied parameters, so "one fixed climatology" holds only for the continuous part of the distribution.
- **Applied fix:** The page now states the limit. Commit `cafcb518`.

#### [P3] SPEI's Pearson failure was described as one exception type — Fixed
- **Location:** `docs/choosing-parameters.md` (distribution section)
- **Origin:** Documentation accuracy, contracts
- **Problem:** Insufficient non-zero values raise `InsufficientDataError`, a failed L-moment fit raises `PearsonFittingError`, and the non-spatial path can substitute zeroed parameters for a failing calendar period with only a high-failure-rate warning. Naming a single type was wrong.
- **Applied fix:** The page states the policy difference that matters (SPI falls back to gamma; SPEI does not) and defers the error contract to `data_requirements.md`. Commit `cafcb518`.

#### [P3] The command-line file-name pattern was presented as universal — Fixed
- **Location:** `docs/writing-outputs.md` (command-line section)
- **Origin:** Documentation accuracy, contracts
- **Problem:** `<base>_<index>_<distribution>_<scale>.nc` is SPI/SPEI only; PNP writes `..._pnp_03.nc`, PET writes `..._pet_thornthwaite.nc`, and the Palmer family uses its own variable names.
- **Applied fix:** The pattern is scoped to SPI and SPEI, with examples for the others. Commit `cafcb518`.

#### [P3] CLI output provenance was over-promised — Fixed
- **Location:** `docs/writing-outputs.md` (verification section)
- **Origin:** API contracts
- **Problem:** The verification section framed the CLI check as provenance verification, but the CLI records only `long_name` and the valid range; distribution, scale, and calibration years are not in the file.
- **Applied fix:** The page now states what the file records and that the run arguments have to be kept alongside it. Commit `cafcb518`.

#### [P3] The Zarr example fails on a rerun — Fixed
- **Location:** `docs/writing-outputs.md` (Zarr section)
- **Origin:** Correctness and edge cases
- **Problem:** `to_zarr` without a `mode` uses `w-` and raises `FileExistsError` on an existing store, while the surrounding text covers reruns. Reproduced by running the block twice.
- **Applied fix:** The page states that the call creates the store and points reruns at the staged swap it already described. Commit `cafcb518`.

#### [P3] The performance page mis-placed the notebook client teardown — Fixed
- **Location:** `docs/performance.md` ("Running on a cluster")
- **Origin:** Tests and verification
- **Problem:** The sentence said the client cell closes the client; the close is in the write cell's `finally` block.
- **Applied fix:** Wording now attributes the close to the write cell. Commit `cafcb518`.

#### [P4] Smaller corrections — Fixed
- **Location:** both new pages
- **Origin:** Multiple lanes
- **Problem:** A recommendation was attributed to `performance.md` that it does not contain; a troubleshooting link pointed at a page with no write-failure content; the `zarr` package (not a runtime dependency) was not listed as a prerequisite; the `--chunksizes input` copy claim was restated although it does not happen with the inputs the CLI opens by default.
- **Applied fix:** Attribution removed, link replaced with the operational-guidance link, `zarr` named as a prerequisite, and the `--chunksizes` sentence replaced with a pointer to the CLI reference. Commit `cafcb518`.

### Remaining blockers

None. Three items are deferred rather than fixed, and none blocks this documentation PR:

1. **`--chunksizes input` is a silent no-op with the default backend.** `_input_chunksizes` (`src/climate_indices/__main__.py`) gates on `da.encoding.get("contiguous", True)`; the h5netcdf backend does not set `contiguous`, so the copy is skipped. Reproduced: `--index pnp --chunksizes input` over a chunked h5netcdf input writes `chunksizes: None`. Docs cannot fix this; it needs the gate's default changed (or `reference.md` caveated) in a separate code PR. Recorded in the PR body.
2. **Dask `fitting_params` block misalignment** (finding above) is now documented as a limitation; the adapter-side fix belongs in its own change.
3. **The Pearson per-calendar-period failure substitution** in the non-spatial path (zeroed parameters plus a high-failure-rate warning) is a behaviour question the guide no longer asserts anything about; it is not a documentation defect and is left to the scientific-behaviour work.

monocongo:
Deferred item 1 from the review note is now tracked as #1080 (`--chunksizes input` is a silent no-op with the h5netcdf backend).

PR #1081, opened 2026-09-20
fix(cli): preserve h5netcdf input chunksizes

### LINKED ISSUE #1080: --chunksizes input is a silent no-op with the h5netcdf backend
**Describe the bug**

The `--chunksizes input` command-line option silently does nothing: the output NetCDF is written with the writer's default (contiguous) layout instead of copying the first chunked input variable's on-disk chunks, as `docs/reference.md` documents.

Root cause: `_input_chunksizes` in `src/climate_indices/__main__.py` gates the copy on

```python
if not da.encoding.get("contiguous", True):
```

The `contiguous` encoding key is not set by the h5netcdf backend, which is the only NetCDF backend installed here, so the default `True` short-circuits every variable and `_input_chunksizes` returns `((), ())`. The `chunksizes` key, which h5netcdf does set for chunked variables, is never consulted.

**To Reproduce**

1. Write a chunked NetCDF input (the CLI expects `(lat, lon, time)` for gridded input):

```python
import numpy as np
import pandas as pd
import xarray as xr

rng = np.random.default_rng(3)
time = pd.date_range("1981-01-01", periods=40 * 12, freq="MS")
ds = xr.Dataset(
    {"prcp": (("lat", "lon", "time"), rng.gamma(2.0, 15.0, (3, 4, time.size)), {"units": "mm"})},
    coords={"time": time, "lat": [25.0, 30.0, 35.0], "lon": [-100.0, -95.0, -90.0, -85.0]},
)
ds.to_netcdf("prcp.nc", encoding={"prcp": {"chunksizes": (2, 4, 24)}}, engine="h5netcdf")
```

2. Run the CLI with `--chunksizes input`:

```bash
climate_indices --index pnp --periodicity monthly --scales 3 \
  --netcdf_precip prcp.nc --var_name_precip prcp \
  --output_file_base out --calibration_start_year 1981 --calibration_end_year 2010 \
  --chunksizes input
```

3. Inspect the output:

```python
import xarray as xr

with xr.open_dataset("out_pnp_03.nc", engine="h5netcdf") as dataset:
    print(dataset["pnp_03"].encoding.get("chunksizes"))
```

Result: `None` (contiguous). Expected: `(2, 4, 24)` — the input variable's chunks, reordered to the output's dimension order.

**Expected behavior**

`--chunksizes input` writes the output with the copied input chunks, as documented, or states the limitation in `docs/reference.md` (and the `--chunksizes` mention in `docs/workflow-examples.md`, if any) if the copy is intentionally limited to backends that report `contiguous`.

**Desktop (please complete the following information):**

- OS: macOS 15 (arm64)
- Python: 3.14.7
- climate_indices: `main` at `180f1715` (3.0.0), `h5netcdf 1.x`, `xarray 2026.1.0`

**Additional context**

- Found while reviewing #1076 (docs how-to guides); reproduced independently of the new documentation, and the docs PR carries a pointer to the reference instead of the claim.
- Suggested minimal fix: default the gate to `False`, i.e. `if not da.encoding.get("contiguous", False):`. The subsequent `chunksizes` check already skips variables whose `chunksizes` is `None` or empty, so treating an unknown `contiguous` as "not contiguous" only acts when chunk sizes are actually present.
- `tests/test_main_chunksizes.py` sets `contiguous=False` in the dataset encoding explicitly, so it passes either way; a regression test would need a real chunked h5netcdf file (the pattern in `tests/test_main_kbdi.py::test_chunked_inputs_keep_time_whole_and_copy_input_chunksizes`).

monocongo:
## Summary
- retain input chunks when h5netcdf omits the `contiguous` encoding key
- cover writing output from a chunked h5netcdf input

Closes #1080

## Validation
- `uv run mypy src/`
- `uv run pytest`

## Summary by Sourcery

Preserve input chunking when generating CLI output from h5netcdf datasets.

Bug Fixes:
- Preserve chunk sizes from h5netcdf inputs when the encoding omits the `contiguous` key.

Tests:
- Add coverage for writing output from a chunked h5netcdf input and retaining its chunk sizes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of chunked input data when chunk metadata is explicitly marked as non-contiguous.
  * Preserved input chunk sizes when generating the PNP index, including files without explicit contiguous metadata.

* **Tests**
  * Added coverage confirming that input chunk sizes are retained for H5netCDF-written datasets.


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/1081
- **Lanes run:** 3 — correctness/edge cases, tests, maintainability
- **Validation:** `uv run ruff check src/ tests/` — passed; `uv run ruff format --check src/ tests/` — passed; `uv run mypy src/` — passed; `uv run pytest` — 2260 passed, 432 deselected
- **Fix commits:** [474f1387 — fix(cli): preserve h5netcdf input chunksizes](https://github.com/monocongo/climate_indices/commit/474f1387fc083aa0887a81cd662345ef9ef9e270)

### Findings and fix status
No additional verified review findings.

### Remaining blockers
None.

monocongo:
- **Review target:** [PR #1081 — `fix(cli): preserve h5netcdf input chunksizes`](https://github.com/monocongo/climate_indices/pull/1081). Reviewed at head `474f1387`; **the PR was merged as `2b4b6490` while this review was running**, so the fix stack below is rebased onto `main` (`1d04cda7`) and is intended as a follow-up, not as changes to this PR.
- **Lanes run:** 7 lane executions over 3 passes — correctness/edge cases (deepseek), test quality (deepseek), public contracts/CLI compatibility (gpt-5.6-terra), docs accuracy (gpt-5.6-terra); two re-validation passes after the first and second fix rounds.
- **Validation:** `ruff check src/ tests/` pass · `ruff format --check src/ tests/` pass (98 files) · `mypy src/` pass (28 files) · `pytest` **2266 passed, 432 deselected** (benchmark + validation markers) · every new guard was re-run with its source fix reverted and failed as expected.
- **Fix commits:** [`#1084`](https://github.com/monocongo/climate_indices/pull/1084) — `bc228229`, `a9c748f1`, `cbd6ff1f`, `0571e1d3`, `7fb1de58`, `528dc0f6`, `dfe39a38`, `4b2f7624` on `fix/1081-chunksize-followups` (base `origin/main` at `1d04cda7`).

### Findings and fix status
#### [P2] Copied input chunks larger than the output dimension were silently dropped — Fixed
- **Location:** `src/climate_indices/__main__.py` (helper `_trimmed_output_encodings` at 762–780; call sites 1596 and 1635)
- **Origin:** Correctness/edge cases; public contracts
- **Problem:** An input written with an unlimited dimension can legally report a chunk larger than the data on disk (e.g. `(2, 3, 100)` for 24 records). h5netcdf silently drops such a chunk size at write time, so `--chunksizes input` still produced a contiguous output — the same silent no-op #1080 is about, one layer down. On the daily path the trim has to use the *written* shape: `_write_single_output` converts 366-day years back to the Gregorian calendar after the chunk sizes are computed, so a chunk such as `(2, 3, 366)` for 365 records (or `(2, 3, 4019)` for 4018) survived any clamp made against the pre-transform shared-array shape and was then dropped.
- **Evidence:** unlimited-`time` inputs with a chunk larger than the record count, `--chunksizes input`: base and the first clamp commit wrote `encoding["chunksizes"] is None` (h5py `chunks=None`) with no warning; `0571e1d3` writes the trimmed tuple and logs `Trimming copied input chunksizes (2, 3, 100) to the output shape (2, 3, 24)`. Data values are unchanged versus base (md5 equal) across monthly, divisions, daily, Palmer and `--index all` runs; only the HDF5 chunk layout differs.
- **Applied fix:** Trim the copied chunk sizes where the data shape is known (`_write_single_output`, `_write_palmer_outputs`) and warn when a size is reduced, commit `0571e1d3` (the earlier clamp, `bc228229`, was correct for monthly/divisions but stale for daily). Regression tests: `tests/test_main_chunksizes.py:119` (monthly) and `:157` (daily), commit `528dc0f6`.

#### [P2] KBDI copied its input chunk sizes uncapped — Fixed
- **Location:** `src/climate_indices/__main__.py` (1888–1900)
- **Origin:** Correctness/edge cases
- **Problem:** The KBDI path copied the precipitation variable's `chunksizes` verbatim (pre-existing, and not changed by the PR itself), so an oversized chunk was silently dropped there too — and it contradicted the contract this stack documents for `--chunksizes input`.
- **Evidence:** daily precipitation chunk `(2, 3, 12000)` over 11000 records → output written contiguous; the added test fails without the fix and passes with it. A lane reproduced the same with `(2, 3, 15000)` over 14610 records.
- **Applied fix:** Route the KBDI copy through the same trim helper, commit `dfe39a38`; guarded by `tests/test_main_kbdi.py:292`.

#### [P3] Which input variable's chunks were copied depended on hash-seed set order — Fixed
- **Location:** `src/climate_indices/__main__.py:871`
- **Origin:** Public contracts/CLI compatibility
- **Problem:** `xr.open_mfdataset(list(set(files)), ...)` destroyed the request order, so with two chunked inputs (precipitation and PET files) the "first chunked input variable" — and therefore the output layout — varied with the per-process string hash seed. Harmless while h5netcdf copied nothing; newly reachable for the inputs this PR enables.
- **Evidence:** two identical SPEI runs under `PYTHONHASHSEED=0` and `=2` wrote chunks `(2, 3, 12)` then `(2, 3, 6)` before the fix; both write `(2, 3, 12)` after it.
- **Applied fix:** `list(dict.fromkeys(files))` keeps the de-duplication and the request order, commit `a9c748f1`; guarded by `tests/test_main_chunksizes.py:196`, which runs both file orderings so a `set`-based regression cannot pass by hash-seed luck.

#### [P3] The documented contract omitted the selection and trimming rules — Fixed (documentation)
- **Location:** `docs/reference.md:81`, `docs/troubleshooting.md:1224`, docstring at `src/climate_indices/__main__.py:675–680`
- **Origin:** Docs accuracy; public contracts
- **Problem:** The docs promised that `input` "copies the on-disk chunks of the first chunked input variable" with no caveat. The first chunked *data* variable can be dimension-incompatible with the output — for example a chunked 1-D `lat` listed before a chunked `(division, time)` precipitation variable — in which case `_reordered_chunksizes` warns and the writer chooses the layout. The docstring's "if no variable is chunked" was also inexact for a variable carrying `contiguous=True` plus chunk sizes.
- **Evidence:** `_input_chunksizes(dataset)` returns `((1,), ('division',))` for a chunked `lat`; `_reordered_chunksizes` then logs `Ignoring '--chunksizes input': chunked variable dimensions ('division',) do not match output dimensions ('division', 'time')` and copies nothing.
- **Applied fix:** Document the eligibility, dimension-match, reorder and trim rules in `docs/reference.md` and `docs/troubleshooting.md` and state them exactly in the docstring (commits `cbd6ff1f`, `7fb1de58`). The dimension-incompatible selection behavior is left unchanged deliberately — picking the first *dimension-compatible* variable instead would change output layouts for existing netCDF4 users and belongs in its own change.

#### [P3] The PR's new test simulated a condition that never occurs — Fixed
- **Location:** `tests/test_main_chunksizes.py:63–97`
- **Origin:** Tests
- **Problem:** The test monkeypatched `open_mfdataset` to `pop("contiguous")` from the opened encoding. h5netcdf never sets that key, so the patch was a no-op that implied the condition was being manufactured; had a backend ever started setting it, the test would silently stop exercising the intended input.
- **Evidence:** a chunked h5netcdf input reports `{'chunksizes': (2, 3, 12), 'fletcher32': False, ...}` — no `contiguous` key; the popped value is `None`.
- **Applied fix:** Replace the no-op with an assertion that pins the premise (`"contiguous" not in opened["prcp"].encoding`), commit `bc228229`.

#### [P3] No coverage for contiguous inputs or for two of the write paths — Fixed
- **Location:** `tests/test_main_chunksizes.py:100`, `tests/test_main_palmers.py:274`, `tests/test_main_kbdi.py:292`
- **Origin:** Tests
- **Problem:** Flipping the `contiguous` default to `False` makes an absent key mean "chunked", so the safe direction — a genuinely contiguous input that must not gain chunks — had no guard. Separately, the Palmer and KBDI call sites of the new trim had no coverage: reverting the Palmer one broke all five output files, and reverting the KBDI one broke KBDI, with the test suite still green.
- **Evidence:** with the `_write_palmer_outputs` trim reverted, `tests/test_main_chunksizes.py` still passed 8/8.
- **Applied fix:** `test_input_chunksizes_ignores_contiguous_inputs` (`bc228229`), the daily trim test (`528dc0f6`), the Palmer writer test and the both-orderings file-order test (`dfe39a38`, `4b2f7624`). Each was re-run with its source fix reverted and failed.

#### [P4] Stale comment on the forced write engine — Blocked (out of scope)
- **Location:** `src/climate_indices/__main__.py:920–921`
- **Origin:** Public contracts
- **Problem:** The comment justifies the `h5netcdf` pin with "the supported xarray versions still default to scipy when netCDF4 is absent", but with the pinned `xarray>=2025.6.1` the default write engine is h5netcdf — every `--chunksizes none` output is already HDF5. Pre-existing, comment only, no behavior impact.
- **Status:** Not fixed: pre-existing text unrelated to this change. Cleanup requires deciding whether the explicit engine pin is still wanted.

### Remaining blockers
None for the reviewed change itself: PR #1081 is merged, and it is sound — the `contiguous` default flip copies chunk sizes for chunked h5netcdf inputs and copies nothing for contiguous ones. The items above are the gaps the review found around it; the fix stack is green on the full suite and every fixed finding has a test that fails without its fix. It is pushed as `fix/1081-chunksize-followups` (rebased on `1d04cda7`) and opened as #1084 for review. Two follow-ups were deliberately left out of scope and are recorded above: selecting the first dimension-compatible variable when the first chunked one cannot be reordered, and the stale engine comment.

PR #1084, opened 2026-09-20
fix(cli): honor --chunksizes input in every output writer

monocongo:
## Summary

Follow-up to the review of #1081 (merged as `2b4b6490`), which fixed
`--chunksizes input` being a silent no-op with the h5netcdf backend. The review
found the same failure one layer down: a copied chunk larger than the output
dimension is dropped by the writer, so the option is still silently ignored for
inputs written with an unlimited dimension.

- trim copied input chunk sizes to the shape actually written
  (`_write_single_output`, `_write_palmer_outputs`, and the KBDI path), warning
  when a size has to be reduced. The daily path needs the post-Gregorian shape,
  since `_write_single_output` shrinks 366-day years back to the calendar after
  the chunk sizes are computed.
- de-duplicate the input file list without reordering it, so the variable whose
  chunks get copied no longer depends on the per-process string hash seed.
- document the eligibility, dimension-match, reorder and trim rules of
  `--chunksizes input`, and make the `_input_chunksizes` docstring exact.
- tests: oversized chunks (monthly, daily, KBDI, Palmer), contiguous inputs, and
  both input-file orderings.

## Validation

- `uv run mypy src/`
- `uv run pytest` — 2266 passed, 432 deselected

Each new test was re-run with its source fix reverted and fails, so the guards
are real.

## Notes

Full findings, evidence and lane attribution are in the review note on #1081.
Not addressed here, deliberately: selecting the first dimension-compatible
variable when the first chunked variable cannot be reordered (a behavior change
for existing netCDF4 users), and the stale comment about xarray's default write
engine.

## Summary by Sourcery

Honor copied input chunk sizes consistently across all CLI output writers while safely adapting them to the output shape.

Bug Fixes:
- Ensure `--chunksizes input` is honored across monthly, daily, KBDI, and Palmer output writers by trimming oversized copied chunks to the written output shape and warning when trimming occurs.

Enhancements:
- Make input-file de-duplication stable so copied chunk selection follows the requested file order.
- Clarify the eligibility and dimension, reorder, and trim behavior of `--chunksizes input` in user documentation and internal API documentation.
- Use the HDF5-backed writer whenever copied input chunk sizes are requested.

Documentation:
- Document how `--chunksizes input` selects, reorders, trims, and ignores incompatible input chunk sizes.

Tests:
- Add coverage for oversized chunks, contiguous inputs, output-shape trimming across writer paths, and stable input-file ordering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `--chunksizes input` so copied chunk sizes are trimmed to match output data shapes.
  * Improved chunk handling for standard index, Palmer, and KBDI outputs, including oversized input chunks.
  * Preserved input-file order when selecting chunk sizes for consistent results.

* **Documentation**
  * Clarified chunk-size selection, reordering, trimming, and behavior when dimensions do not match.
  * Updated troubleshooting guidance with the refined behavior and KBDI example.


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/1084
- **Lanes run:** 13 successful lane runs — correctness/edge cases (4), API/contracts (2), tests (4), maintainability/docs (3), including reruns after the base update and fix pass
- **Validation:** `uv run ruff check src/ tests/` passed; `uv run ruff format --check src/ tests/` passed (98 files); `uv run mypy src/` passed (28 files); `uv run pytest -q` passed (2273 passed, 432 deselected)
- **Fix commits:** `ca229c81` — test(cli): pin contiguous chunk check to h5netcdf; `9c6e0e41` — test(cli): cover de-duplicated shared input files; `91edb36f` — fix(cli): warn once when trimming Palmer chunks; `7037f73b` — docs(cli): correct chunk-writer rationale; `60d13f96` — test(cli): strengthen copied chunk trim guards; `cf4b74d8` — test(cli): scan past contiguous input variables

### Findings and fix status
#### P3 Engine-dependent contiguous-input assertion — Fixed
- **Location:** `tests/test_main_chunksizes.py` (lines 100–113)
- **Origin:** Correctness and tests personas
- **Problem:** The test asserted h5netcdf-specific encoding metadata through xarray's default backend. Environments with netCDF4 installed could fail although `_input_chunksizes` remained correct.
- **Applied fix:** Pin the read to `engine="h5netcdf"`; commit `ca229c81`.

#### P3 Shared-file de-duplication lacked a regression guard — Fixed
- **Location:** `tests/test_main_chunksizes.py` (lines 201–261)
- **Origin:** Tests persona
- **Problem:** Existing cases used two distinct files, so removing de-duplication still passed. Supplying one file for precipitation and PET twice makes `open_mfdataset` fail during combination.
- **Applied fix:** Add a shared-file case and assert the path reaches `open_mfdataset` once; commit `9c6e0e41`.

#### P3 Palmer chunk trimming emitted five identical warnings — Fixed
- **Location:** `src/climate_indices/__main__.py` (lines 1648–1665)
- **Origin:** Maintainability/docs persona
- **Problem:** The five-output Palmer loop recalculated the same trimmed encoding and logged the same warning once per output.
- **Applied fix:** Compute the shared encoding once before the loop and assert one warning; commit `91edb36f`.

#### P3 Trim tests did not prove fitting axes were preserved — Fixed
- **Location:** `tests/test_main_chunksizes.py` (lines 124–158)
- **Origin:** Tests persona
- **Problem:** Every oversized fixture expected chunks equal to the output shape, so replacing all copied chunks with the shape would pass instead of trimming only oversized axes.
- **Applied fix:** Assert `(1, 2, 100)` becomes `(1, 2, 24)` for output shape `(2, 3, 24)`; commit `60d13f96`.

#### P3 Contiguous-variable test did not prove scanning continued — Fixed
- **Location:** `tests/test_main_chunksizes.py` (lines 100–121)
- **Origin:** Tests persona
- **Problem:** A single-variable fixture could not distinguish skipping a contiguous variable from stopping the scan, allowing a later eligible variable to be silently ignored.
- **Applied fix:** Add a later chunked variable and assert its chunks and dimensions are selected; commit `cf4b74d8`.

#### P4 Chunk-writer comments described obsolete backend selection — Fixed
- **Location:** `src/climate_indices/__main__.py` (lines 950–952 and 1916–1919)
- **Origin:** Correctness persona
- **Problem:** Comments claimed supported xarray versions default to scipy when netCDF4 is absent, but required h5netcdf precedes scipy in backend selection.
- **Applied fix:** State the actual reason: explicitly pin the HDF5-backed writer so copied chunks are honored; commit `7037f73b`.

#### P4 No test guarded the no-trim warning path — Fixed
- **Location:** `tests/test_main_chunksizes.py` (lines 63–98)
- **Origin:** Tests persona
- **Problem:** An unconditional trim warning would leave the suite green for already-fitting chunks.
- **Applied fix:** Assert fitting copied chunks emit no trim warning; commit `60d13f96`.

### Remaining blockers
None.

PR #1088, opened 2026-09-21
fix(xarray): fail closed when a declared periodicity cannot resolve

### LINKED ISSUE #759: _resolve_periodicity fails open, silently disabling xarray calendar conversion
> *This was generated by AI during triage.*

## Summary

`_resolve_periodicity` in `src/climate_indices/xarray_adapter.py` (line 321) returns `None` whenever it cannot confidently determine a periodicity. Its caller treats `None` as "no calendar plan, no validation", so every uncertain case silently disables the mechanism whose entire purpose is to prevent silently wrong results.

This is not currently reachable — see below — but it is a fail-open default inside a safety check, and it is one signature change away from mattering.

## Evidence

```python
def _resolve_periodicity(func, modified_args, modified_kwargs, inferred_params):
    periodicity = inferred_params.get("periodicity")
    if periodicity is not None:
        return periodicity if isinstance(periodicity, compute.Periodicity) else None

    try:
        bound = inspect.signature(func).bind_partial(*modified_args, **modified_kwargs)
    except TypeError:
        return None

    periodicity = bound.arguments.get("periodicity")
    return periodicity if isinstance(periodicity, compute.Periodicity) else None
```

Three separate paths return `None` for reasons that are not "this function has no periodicity":

1. `inferred_params["periodicity"]` is present but not a `Periodicity` instance
2. `bind_partial` raises `TypeError` on an argument mismatch
3. the bound `periodicity` argument is present but not a `Periodicity` instance

`_resolve_daily_calendar_plan` then returns `None`, and `_compute_with_daily_calendar_plan` runs the NumPy core against unconverted Gregorian values — the exact drift #757 removed.

**Confirmed not exploitable today.** A string periodicity such as `spi(..., periodicity="daily")` is rejected downstream by `indices.spi` with `InvalidArgumentError` before any result is returned, so no drifted output escapes to a caller. The protection is incidental to this function, not provided by it.

## Current behavior

An unresolvable periodicity is indistinguishable from a function that legitimately has no `periodicity` parameter (`pci`, and any wrapper without one). Both produce `None`, and neither is logged.

## Desired behavior

Distinguish "this function takes no periodicity" from "this function takes a periodicity and I could not resolve it". The first is normal and should stay quiet. The second should not silently skip calendar conversion — at minimum it should emit a structured log event, and preferably raise, since a wrapped function that declares a `periodicity` parameter whose value cannot be resolved indicates a programming error at the seam rather than a user input problem.

Follow the module's existing conventions: raise from the `ClimateIndicesError` hierarchy in `exceptions.py`, never a bare exception, and use `get_logger()` rather than stdlib logging. Add tests covering each of the three paths above.

Do not weaken the downstream `InvalidArgumentError` validation in `indices.py` — this issue is about the seam failing closed, not about relocating existing checks.

## Context

Follow-up from the code review on #757. Flagged as advisory rather than blocking because it is unreachable through the current public API.

monocongo:
## Summary

`_resolve_periodicity` — the xarray seam that decides whether daily calendar conversion is needed — returned `None` for three paths that do not mean "this function has no `periodicity` parameter":

- `inferred_params["periodicity"]` is present but not a `Periodicity`
- `bind_partial` raises `TypeError` on an argument mismatch
- the bound `periodicity` argument is present but not a `Periodicity`

`_resolve_daily_calendar_plan` treats `None` as "no calendar plan", so each path silently disabled the daily 366-day conversion. This is not exploitable through today's public API (`indices.spi` rejects a string periodicity downstream), but the check whose job is to prevent silently wrong output was failing open.

`_resolve_periodicity` now returns `None` only when the wrapped function does not declare a `periodicity` parameter. Every other unresolved case emits a `periodicity_resolution_failed` structured log event and raises `PeriodicityError` (the `InvalidArgumentError` specialization for periodicity values), so a declared-but-unresolvable periodicity can no longer skip calendar conversion. A non-`Periodicity` value keeps the core validators' `Invalid periodicity argument` message, so the documented troubleshooting error is unchanged for xarray callers. Declared defaults are applied after binding, so a custom adapter function declaring `periodicity=Periodicity.monthly` resolves to that default instead of failing closed.

The downstream `InvalidArgumentError`/`PeriodicityError` validation in `indices.py` is unchanged, per the issue.

## Scope decision

The binding-failure branch is unit-covered but not reachable through the decorator: `_infer_temporal_parameters` swallows its own `bind_partial` `TypeError` and infers every temporal parameter, which then overwrites the caller's explicit values. Fixing that is a separate behavior change (what a call whose arguments do not bind should do), so it is deferred to #1090; this PR does not attempt it.

## Tests

`tests/test_xarray_adapter.py`:

- `TestResolvePeriodicity` covers each of the three fail-open paths, the quiet `None` for functions without a `periodicity` parameter, pass-through of a resolved `Periodicity` (inferred and explicit), a declared default, and an explicit `None`.
- `TestXarrayAdapterIntegration::test_invalid_periodicity_fails_closed_before_compute` pins that an invalid explicit periodicity raises at the seam with the wrapped function body never executing.
- `TestXarrayAdapterIntegration::test_declared_default_periodicity_is_honored` pins the default-resolution regression fix.

## Verification

- `uv run mypy src/` — pass (28 source files)
- `uv run pytest` — 2283 passed, 432 deselected

Closes #759

## Summary by Sourcery

Make xarray periodicity resolution fail closed so invalid declared values cannot bypass calendar conversion.

Bug Fixes:
- Prevent declared but invalid or unresolvable periodicity values from silently disabling daily calendar conversion.
- Raise PeriodicityError and emit structured diagnostics when periodicity resolution fails.

Enhancements:
- Honor declared default periodicity values while distinguishing functions that do not declare periodicity parameters.

Tests:
- Add unit and integration coverage for periodicity resolution failures, defaults, explicit values, and prevention of wrapped-function execution on invalid input.


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/1088 — branch `fix/759-periodicity-fail-closed`, head `da2774a8`
- **Lanes run:** 4 — correctness/edge cases (deepseek-flash), xarray dataflow (deepseek-flash), public API contracts/compatibility (gpt-5.6-terra), tests (deepseek-flash)
- **Validation:** `uv run ruff check src/ tests/` pass; `uv run ruff format --check src/ tests/` pass; `uv run mypy src/` pass (28 source files); `uv run pytest` 2283 passed, 432 deselected
- **Fix commits:** [da2774a8 — fix(xarray): honor declared periodicity defaults and canonical error message](https://github.com/monocongo/climate_indices/commit/da2774a834288ff9d8d3c31bada0c21d3a7856ed)

### Findings and fix status

#### [P2] Defaulted periodicity raised instead of resolving — Fixed
- **Location:** `src/climate_indices/xarray_adapter.py` (`_resolve_periodicity`, binding path)
- **Origin:** correctness, dataflow, contracts, and tests lanes (deduplicated; same underlying claim)
- **Problem:** `Signature.bind_partial` does not apply defaults, while `_infer_temporal_parameters` counts a defaulted parameter as provided and skips inference. A wrapped function declaring `periodicity=Periodicity.monthly` therefore hit the new fail-closed branch on every xarray call (`PeriodicityError: ... got None`). Reproduced through the decorator; the base revision returned the computed result, so this was a regression introduced by the first commit of this PR.
- **Applied fix:** `bound.apply_defaults()` after binding, matching `_infer_temporal_parameters`. Regression tests added at unit level (`TestResolvePeriodicity::test_declared_default_is_returned`) and decorator level (`test_declared_default_periodicity_is_honored`). A default of `None` still fails closed and is pinned by `test_explicit_none_raises`. Commit `da2774a8`.

#### [P3] Adapter message diverged from the documented `Invalid periodicity argument` error — Fixed
- **Location:** `src/climate_indices/xarray_adapter.py` (non-`Periodicity` raise); documented in `docs/troubleshooting.md` and `docs/deprecations/api-changes.md`
- **Origin:** contracts lane
- **Problem:** the seam-specific message changed what `indices.spi(precip_da, ..., periodicity="monthly")` reports for an xarray input, while the troubleshooting guide and the 3.0.0 compatibility note document the core validators' `Invalid periodicity argument` message as unchanged.
- **Applied fix:** the adapter raises the same canonical message and `periodicity_value` as `indices._validate_periodicity`, so the documented error still matches for xarray callers. The integration test now matches `Invalid periodicity argument`. Commit `da2774a8`.

#### [P2] `_infer_temporal_parameters` masks binding failures and overrides explicit parameters — Deferred
- **Location:** `src/climate_indices/xarray_adapter.py` (`_infer_temporal_parameters`, `except TypeError` branch)
- **Origin:** dataflow lane
- **Problem:** on `bind_partial` failure, the helper sets `provided_params = set()`, infers every temporal parameter, and `call_kwargs.update(inferred_params)` then overwrites the caller's explicit values. A typo'd keyword makes a call silently compute with inferred year/periodicity instead of raising (reproduced). This is pre-existing on `origin/main`; it also makes the new binding-failure branch unreachable through the decorator.
- **Status:** not fixed in this PR. Changing what a call whose arguments do not bind should do is a separate behavior decision outside #759's seam, and the fix would newly expose a binding failure as a periodicity-specific error, which needs its own design. The new branch remains correct and unit-covered if that inference is fixed. Tracked as #1090.

### Remaining blockers
None. The deferred finding is pre-existing and tracked in #1090; it is not required for this PR's scope.

Issue #1092, opened 2026-09-21
xarray_adapter: unknown keyword arguments are silently dropped instead of raising TypeError


monocongo:
**Describe the bug**

`xarray_adapter` filters call kwargs through the wrapped function's signature (`valid_kwargs`) before calling it. An unknown keyword argument — a typo, for example — is dropped rather than raising `TypeError`, so the call silently computes without it.

This is the residual half of the problem tracked by #1090. #1090 fixed the more severe symptom (a bind failure let inference fabricate temporal parameters and override the caller's explicit values); with that fix, a call whose arguments do not bind no longer gets inferred values, but for a wrapped function that does not declare `periodicity` there is no remaining seam that rejects the call, and the unknown keyword is still discarded.

**To Reproduce**

```python
@xarray_adapter()
def needs_start_year(values, data_start_year):
    ...

# monthly DataArray, 1980-2019
needs_start_year(da, data_start_year=1990, bogus=1)
```

Expected: `TypeError: needs_start_year() got an unexpected keyword argument 'bogus'` (or an explicit adapter error naming the argument).

Observed: computes with `data_start_year=1990`; `bogus` never reaches the wrapped function and nothing raises.

**Additional context**

- `src/climate_indices/xarray_adapter.py` builds `valid_kwargs` from `call_kwargs` in both the Dask and in-memory paths; unknown keys are silently excluded.
- The typed public API (`typed_public_api.py`) already binds against the wrapped signature and rejects unknown keywords, so this affects direct/custom `@xarray_adapter()` usage.
- Raised by the #1090 review lanes and deliberately left out of that fix's scope, whose minimum requirement was to stop overriding explicit values.

PR #1093, opened 2026-09-21
fix(xarray): skip temporal inference when arguments do not bind

### LINKED ISSUE #1090: xarray_adapter: a bind failure in _infer_temporal_parameters silently overrides explicit temporal params
`_infer_temporal_parameters` catches the `TypeError` from `Signature.bind_partial` and continues with `provided_params = set()`:

```python
try:
    bound = sig.bind_partial(*modified_args, **modified_kwargs)
    bound.apply_defaults()
    provided_params = set(bound.arguments.keys())
except TypeError:
    # if binding fails, skip inference
    provided_params = set()
```

The comment says "skip inference", but the code does the opposite: with `provided_params` empty, every temporal parameter is inferred and the decorator's `call_kwargs.update(inferred_params)` then overwrites the caller's explicit values. An argument that does not bind — for example a typo'd keyword — therefore downgrades a hard validation error into silently computing with inferred parameters.

## Evidence

```python
@xarray_adapter()
def masked(values, data_start_year, periodicity):
    return np.full(values.shape, float(data_start_year))

# monthly DataArray, 2000-2002
masked(da, data_start_year=1990, periodicity=compute.Periodicity.daily, bogus=1)
```

Observed: returns a result computed with `data_start_year=2000` and `periodicity=monthly` (both inferred), silently ignoring the explicit arguments and the unknown keyword. Expected: the binding failure surfaces instead of being replaced by inferred values.

This also makes the `PeriodicityError` path added for #759 unreachable through the decorator: whenever `bind_partial` fails, inference fills `periodicity` first, so `_resolve_periodicity` never reaches its binding-failure branch. (The #759 branch is still covered at unit level and remains correct if inference is fixed.)

## Desired behavior

Honor the existing comment: on a binding failure, skip inference (return the params inferred so far, normally none) so the caller's arguments are either used as given or the wrapped function raises for the missing/unexpected arguments. The inference helper should not fabricate temporal parameters for a call whose arguments do not bind.

Surfacing the real `TypeError` would be even better than the adapter's periodicity-specific seam error for cases with no periodicity parameter, but the minimum requirement is to stop overriding explicit values.

## Tests

- Repro above: with a typo'd keyword and explicit `periodicity=daily` on a monthly coordinate, assert the call does not compute with inferred monthly values.
- Normal inference paths must stay green (`tests/test_xarray_adapter.py::TestInferTemporalParameters`).

Found while fixing #759 (see PR #1088).

monocongo:
## Summary

`_infer_temporal_parameters` swallowed the `bind_partial` `TypeError` and continued with `provided_params = set()`, so every temporal parameter was treated as missing and inferred. The decorator then merged those inferred values over the call kwargs (`call_kwargs.update(inferred_params)`), so a call whose arguments did not bind — a typo'd keyword, for example — silently computed with fabricated `data_start_year`/`periodicity` instead of surfacing the error. It also made the `PeriodicityError` binding-failure branch added for #759 unreachable through the decorator, because inference always produced a periodicity first.

The helper now returns its (empty) inference result on a bind failure, honoring its existing "skip inference" comment. Explicit arguments are no longer overridden, and `_resolve_periodicity` reaches its binding-failure branch: a decorated function that declares `periodicity` raises `PeriodicityError` naming the binding failure before computing.

## Scope decision

The issue states that surfacing the real `TypeError` for an unknown keyword would be even better, but sets the minimum at not overriding explicit values. Rejecting unknown keywords in the adapter's `valid_kwargs` filter is pre-existing behavior with a different blast radius (custom adapters, functions without a `periodicity` parameter), so it is deliberately deferred to #1092 (`xarray_adapter: unknown keyword arguments are silently dropped instead of raising TypeError`).

## Tests

`tests/test_xarray_adapter.py`:

- `TestInferTemporalParameters::test_bind_failure_skips_inference` — a genuinely non-binding call gets `{}` instead of inferred params (fails on the previous behavior).
- `TestInferTemporalParameters::test_declared_default_is_not_inferred` — pins `bound.apply_defaults()` so declared defaults keep counting as provided.
- `TestXarrayAdapterIntegration::test_bind_failure_does_not_override_explicit_temporal_params` — the issue's repro: explicit `data_start_year=1990`/`periodicity=daily` plus a typo'd keyword raises at the seam instead of computing with inferred monthly values, and the wrapped function never runs.
- `TestXarrayAdapterIntegration::test_bind_failure_keeps_explicit_params_without_periodicity` — for a function without a `periodicity` parameter, the explicit `data_start_year` reaches the function unchanged.

## Verification

- `uv run mypy src/ tests/test_type_checking.py` — pass (29 source files)
- `uv run pytest` — 2304 passed, 432 deselected

Closes #1090

## Summary by Sourcery

Skip temporal inference when adapter arguments fail signature binding so invalid calls cannot silently execute with inferred values.

Bug Fixes:
- Prevent temporal parameter inference from overriding explicit arguments when call arguments do not bind to the decorated function signature.
- Surface periodicity binding failures instead of allowing invalid calls to proceed with fabricated temporal values.

Enhancements:
- Preserve declared default temporal parameters as provided during inference.

Tests:
- Add coverage for binding failures, declared defaults, and preservation of explicit temporal parameters with and without a periodicity parameter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved temporal parameter handling when invalid, unknown, or duplicate arguments are provided.
  * Prevented inferred dates and periodicity values from replacing explicitly supplied parameters.
  * Preserved valid explicit parameters even when another argument fails validation.
  * Improved error reporting for arguments that do not match a wrapped function’s signature.
  * Prevented invalid arguments from triggering misleading inferred values or unintended execution.


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/1093 (`fix/1090-bind-failure-skip-inference` against `main`)
- **Lanes run:** 3 — correctness/edge cases (deepseek), tests (deepseek), public API contracts/compatibility (openai-codex)
- **Validation:** `uv run ruff check src/ tests/` pass; `uv run ruff format --check src/ tests/` pass; `uv run mypy src/ tests/test_type_checking.py` pass; `uv run pytest` 2304 passed, 432 deselected
- **Fix commits:** [cb93376c — fix(xarray): skip temporal inference when arguments do not bind](https://github.com/monocongo/climate_indices/commit/cb93376c), [5d30df89 — test(xarray): pin bind-failure parameter preservation](https://github.com/monocongo/climate_indices/commit/5d30df89)

### Findings and fix status

#### [P3] Integration test only proved the seam raises, not that explicit params survive — Fixed
- **Location:** `tests/test_xarray_adapter.py` (lines 944–967)
- **Origin:** Tests lane
- **Problem:** The repro test passed because `_resolve_periodicity` raises `PeriodicityError` on the non-binding kwargs; it did not directly assert parameter preservation, and a wrapped function without a `periodicity` parameter could still have its explicit `data_start_year` overridden without any test noticing.
- **Applied fix:** Added `test_bind_failure_keeps_explicit_params_without_periodicity` (line 969), which fails on the pre-fix behavior (`calls == [1980]`) and asserts `calls == [1990]`. Commit [5d30df89](https://github.com/monocongo/climate_indices/commit/5d30df89).

#### [P3] Declared defaults were not pinned as "provided" — Fixed
- **Location:** `src/climate_indices/xarray_adapter.py:1052` (`bound.apply_defaults()`)
- **Origin:** Tests lane
- **Problem:** Removing `apply_defaults()` would let inference override a declared default without any test failing, because the existing defaulted-parameter cases use `periodicity=monthly`, which equals coordinate inference on the monthly fixture.
- **Applied fix:** Added `TestInferTemporalParameters::test_declared_default_is_not_inferred` (line 791). Commit [5d30df89](https://github.com/monocongo/climate_indices/commit/5d30df89).

#### [P3] Unknown keywords silently dropped when the wrapped function has no `periodicity` parameter — Blocked (out of scope)
- **Location:** `src/climate_indices/xarray_adapter.py:1489` and `:1566` (`valid_kwargs` filters)
- **Origin:** Correctness lane
- **Problem:** With no `periodicity` parameter, no seam rejects a non-binding call; an unknown keyword is filtered out before the wrapped function runs, so the call still computes silently (pre-existing behavior, not a regression from this change).
- **Applied fix:** **Blocked:** deliberately outside #1090's stated minimum (stop overriding explicit values). Follow-up tracked by #1092.

### Remaining blockers
None.

PR #1094, opened 2026-09-21
test(benchmark): use absolute xarray overhead budget

### LINKED ISSUE #758: Replace ratio-based xarray overhead budgets with an absolute fixed-cost gate
> *This was generated by AI during triage.*

## Summary

`tests/test_benchmark_overhead.py` gates xarray adapter overhead as a **ratio** against the equivalent NumPy path. That ratio is the wrong shape for the cost it now guards: the calendar validation added in #757 is a fixed per-call cost, so the measured percentage is governed mostly by how cheap the NumPy baseline is, not by whether the adapter regressed.

This is the third operation-specific budget in the file. Each was widened for the same underlying reason, which suggests the metric rather than the numbers is the problem.

## Evidence

Current constants in `tests/test_benchmark_overhead.py`:

- `_OVERHEAD_THRESHOLD = 0.80` (line 35) — shared default
- `_PET_HARGREAVES_OVERHEAD_THRESHOLD = 1.00` (line 41) — widened in #740 for runner variance
- `_PET_THORNTHWAITE_OVERHEAD_THRESHOLD = 1.20` (line 50) — widened in #757

The Thornthwaite widening came from a single `xr.infer_freq` call added by `_build_daily_calendar_plan`, measured locally at **~0.14 ms** against a **~0.4 ms** NumPy baseline. That moved the ratio from comfortably under 80% to 85–98% across repeated local runs, with no change to the adapter's per-element work.

At 1.20 the gate now permits roughly **2.2x** the NumPy path before failing, which is weak protection against a genuine regression. Meanwhile the operations with the heaviest NumPy baselines are held to 0.80, so the same absolute regression is judged far more strictly on a cheap operation than an expensive one — the opposite of what the gate should do.

## Current behavior

`_assert_overhead_within_budget` computes `(xarray_time - numpy_time) / numpy_time` and compares it to a per-operation ratio. Because the adapter's fixed costs (coordinate validation, frequency inference, metadata construction) do not scale with the wrapped computation, the ratio inflates as the baseline shrinks. Every time that inflation crosses a threshold, the response has been to widen that operation's budget.

## Desired behavior

Guard the invariant that actually matters: the adapter's **fixed overhead per call** should stay within an absolute budget (e.g. under ~0.5 ms), independent of how cheap or expensive the wrapped NumPy computation is. A ratio check may still be worth keeping for operations where the per-element cost dominates, but it should not be the only signal, and it should not require a new constant each time a fixed cost is added.

Apply the change consistently across all measured operations rather than to Thornthwaite alone, so the file stops accumulating one-off budgets. Retain the existing diagnostic output — both timings and the budget — on failure, as #740 established.

Do not address this by removing the calendar validation from `pet_thornthwaite`; that check is what prevents a daily coordinate being silently reshaped into 12-month rows. Do not simply raise `_OVERHEAD_THRESHOLD` for everything.

## Context

Follow-up from #757, which made calendar semantics explicit at the xarray seam. Related to #740, which widened the Hargreaves budget for a different cause (GitHub-hosted runner variance) but the same underlying metric weakness.

monocongo:
## Summary

- replace operation-specific xarray/NumPy overhead ratios with one 3 ms absolute fixed-cost budget
- alternate NumPy-first and xarray-first timing pairs, then average each order's median delta to remove order bias
- retain failure diagnostics for overhead, budget, and both path timings
- amend ADR-0004 to record the budget-policy change

## Validation

- `uv run mypy src/`
- `uv run pytest` — 2302 passed, 432 deselected
- `uv run pytest tests/test_benchmark_overhead.py -m benchmark --benchmark-enable -q` — 13 passed, 4 deselected

Closes #758

## Summary by Sourcery

Replace operation-specific xarray overhead ratios with a shared absolute budget and more robust order-balanced benchmark measurements.

Enhancements:
- Standardize xarray adapter performance checks on a shared 3 ms absolute fixed-cost overhead budget.
- Improve benchmark stability by alternating NumPy and xarray measurement order and averaging order-specific median timing deltas.
- Preserve detailed benchmark failure diagnostics and validate the revised measurement policy with deterministic tests.

Documentation:
- Amend ADR-0004 to document the transition from operation-specific overhead ratios to a shared absolute budget.

Tests:
- Update overhead benchmarks and policy tests for the shared fixed-cost budget and order-neutral timing measurements.

## Summary by CodeRabbit

* **Documentation**
  * Updated benchmark overhead guidance to use a shared absolute budget rather than operation-specific ratios.

* **Tests**
  * Standardized xarray adapter overhead checks around a shared 3 ms fixed-cost budget.
  * Improved timing measurements with warmups, alternating execution order, and median-based results.
  * Added validation for invalid trial counts and expanded coverage of the shared budget policy.


monocongo:
- **Review target:** https://github.com/monocongo/climate_indices/pull/1094
- **Lanes run:** 3 — correctness/edge cases, performance/flakiness, requirements/documentation
- **Validation:** `uv run ruff check src/ tests/`, `uv run ruff format --check src/ tests/`, `uv run mypy src/`, and `uv run pytest` passed (2302 passed, 432 deselected); `uv run pytest tests/test_benchmark_overhead.py -m benchmark --benchmark-enable -q` passed (13 passed, 4 deselected)
- **Fix commits:** [c68c695b — test(benchmark): stabilize absolute overhead measurement](https://github.com/monocongo/climate_indices/commit/c68c695bc1afbb075eefaceb4de55e827cd9175e), [358036c0 — test(benchmark): balance overhead trial ordering](https://github.com/monocongo/climate_indices/commit/358036c0cd386e25cd6436e2d3e241db6872dcc2), [09d6b484 — docs(adr): mark xarray calendar record amended](https://github.com/monocongo/climate_indices/commit/09d6b48443cd6de766df5ca2d66aa983bfd1f015), [b87c8fde — test(benchmark): cover order-neutral overhead estimate](https://github.com/monocongo/climate_indices/commit/b87c8fde89b3e6b24365ce766dff6983c2bc460e), [21ced48a — test(benchmark): distinguish paired median estimator](https://github.com/monocongo/climate_indices/commit/21ced48a93ece039a3da6c934d9abae91b7e53b3)

### Findings and fix status
#### P1 Hosted-runner budget below unchanged overhead — Fixed
- **Location:** `tests/test_benchmark_overhead.py` (lines 29–33)
- **Origin:** Performance/flakiness
- **Problem:** Initial 0.75 ms budget would fail unchanged hosted-runner behavior; run 35595995006 measured a 1.785 ms SPEI delta, and recent artifact means reached about 2.5 ms.
- **Applied fix:** Set one 3 ms absolute budget with hosted-runner headroom in commit `c68c695b`.

#### P2 Fixed-order timing could mismeasure adapter overhead — Fixed
- **Location:** `tests/test_benchmark_overhead.py` (lines 344–374)
- **Origin:** Performance/flakiness, correctness/edge cases
- **Problem:** Independent or unevenly ordered timing phases could turn host-speed drift into false failures or hide regressions.
- **Applied fix:** Alternate NumPy-first/xarray-first pairs, balance both orders, and average each order's median delta in commits `c68c695b` and `358036c0`.

#### P3 ADR amendment provenance missing — Fixed
- **Location:** `docs/adr/0004-xarray-calendar-semantics.md` (lines 3–5)
- **Origin:** Requirements/documentation
- **Problem:** Budget-policy text changed while ADR status remained `Accepted`, contrary to the repository's ADR amendment convention.
- **Applied fix:** Marked ADR-0004 amended and cited issue #758 in commit `09d6b484`.

#### P3 Estimator policy test allowed incorrect implementations — Fixed
- **Location:** `tests/test_benchmark_overhead.py` (lines 113–153)
- **Origin:** Correctness/edge cases
- **Problem:** Initial deterministic timings did not distinguish the required per-order median-delta estimator from pooled, mean, minimum, or marginal-median alternatives.
- **Applied fix:** Added asymmetric timing data that uniquely pins balanced per-order median behavior in commits `b87c8fde` and `21ced48a`.

### Remaining blockers
None.

